PR #181, opened 2026-02-22
Add data entities

FH-Inway:
Well, this was fun 😄 

I added/extended metadata extraction for data entities. My current use case is to use it for documenting data entities, which also showcases the recently added label tools.
This also adds EDT metadata and splits the get_enum_info tool into two tools (one for enums and one for edts).
Increases the metadata database by about 100 MB.

From one of my tests, I asked it to create the following prompt template to get information about data entities and their key fields in English and German.

> List the data entities for [USE CASE] in D365FO.
For each entity, return:
>- Entity name
>- Entity label ID and translations (en-US, de)
>- Key fields
> 
> For each key field: 
>- field label ID from entity metadata (if present), otherwise fallback to mapped source field/EDT label ID
>- en-US and de translations for each key-field label ID.
>
> If a key field has no explicit entity label ID, mark it as - and show the fallback label resolution path (for example EDT A -> EDT B -> @LabelFile:LabelId).
Output as a compact table with columns: Entity | KeyField | EntityFieldLabelId | FallbackLabelId | en-US | de.
Use only existing labels; do not infer translations.
If no German translation exists, return de: \<missing\>.

### DIFF
--- a/src/server/mcpServer.ts
+++ b/src/server/mcpServer.ts
@@ -881,7 +881,7 @@ Examples:
         },
         {
           name: 'get_edt_info',
-          description: `📊 Get complete Extended Data Type (EDT) definition including base type, labels, reference table, and EDT properties.
+          description: `📊 Get complete Extended Data Type (EDT) definition including base type, labels, reference table, and EDT properties. EDT names are UNIQUE ACROSS ALL MODELS.
 
 Returns:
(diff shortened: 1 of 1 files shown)


dynamics365ninja (approved):
Great idea! Thanks for that. It needs to be documented as well, leave it to me for this time. If you have any other ideas, please do it.

PR #483, opened 2026-05-18
fix(applyObjectPrefix): handle dot-notation model-name suffixes corre…

laeliand:
## Summary

- SPECIAL CASE A was gated on `endsWith('extension')`, so names like
  `SalesOrderHeaderV4Entity.Contoso` (as Visual Studio generates them)
  fell through to NORMAL CASE and got the prefix prepended to the whole
  base name, producing `ConSalesOrderHeaderV4Entity.Contoso`.
- Expanded the guard to cover all dot-notation names, split into three sub-cases:
  - A1 — suffix already starts with the infix → return as-is (no double-prefix)
  - A2 — suffix ends with `extension` but has a different infix → normalize to `{infix}Extension`
  - A3 — model-name suffix as Visual Studio generates it → return as-is

## Test plan

- [x] `npm test -- tests/utils/objectSuffix.test.ts` — 12/12 pass
- [x] Full suite `npm test` — 444/444 pass
- [x] `npm run build` — clean, no TypeScript errors

### DIFF
--- a/src/utils/modelClassifier.ts
+++ b/src/utils/modelClassifier.ts
@@ -183,31 +183,31 @@ export function applyObjectPrefix(objectName: string, prefix: string): string {
     : prefix.charAt(0).toUpperCase() + prefix.slice(1);   // normalize PascalCase
 
   // SPECIAL CASE A: Dot-notation extension elements — BaseElement.Suffix
-  // Visual Studio names extensions as BaseObject.ModelName (e.g. SalesOrderHeaderV4Entity.Contoso).
-  // Covers three sub-cases:
-  //   A1: suffix already starts with the extension infix → return as-is (no double-prefix)
-  //       e.g. "CustTable.ConExtension" with infix "Con" → "CustTable.ConExtension"
-  //   A2: suffix ends with "extension" but has a different infix → normalize to {infix}Extension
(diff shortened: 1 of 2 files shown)


dynamics365ninja:
Thanks for finding the bug, I'm happy about it. I've also implemented a few changes that resulted from the audit so that it doesn't disrupt the overall operation of the MCP server. See if it's OK for you. Thanks

laeliand:
Thanks thats fine.

laeliand:
@dynamics365ninja I had to rewrite the commit and I unintendedly destroyed your commit in the progress. Sorry about that. Can you please re-apply those additions.

dynamics365ninja:
It got a little complicated, but I finally got it back in. Check that it's functionally OK. Thanks

laeliand:
@dynamics365ninja  Your approach is simpler and correct. The tests look good too. 

I have a flawed merge commit (the original commit) with a references to a project that I am working on, that I dont want public. Can you do squash commit when you merge to main so I can get rid of that commit comment? 

I cant force rewrite it without losing your commits or it is too complicated and I dont want to risk of losing this all again. Much appreciated, thanks.-

laeliand:
ah ha, you already fixed it. thanks ! :)

PR #509, opened 2026-06-01
feat(build): full-build flag, BP check flag-style fix, schema update

laeliand:
build_d365fo_project:
- Add fullBuild parameter: omits -incremental so xppc recompiles all elements from scratch. Fixes stale symbol errors after partial builds.
- Add buildReferencedModels parameter: reads <ModuleReferences> from the model descriptor, resolves custom/ISV dependencies via DFS topological sort (skips Microsoft standard models), and builds each in queue order.
- Add -verbose flag to xppc args: surfaces metadata loading errors (missing assembly refs, XML issues) otherwise silent in standalone mode.
- Stale symbol detection: scans xppc -log output and appends a fullBuild:true tip when stale symbol messages are detected.
- Full log on failure: returns head+tail (up to 300 lines) instead of tail-only so the top of long error logs is never lost.
- Bugfix: fullBuild:true with an existing completed incremental state now discards the stale cache and starts a fresh full build instead of returning the old incremental result.
- Queue-aware state: state file keyed by targetModel throughout a multi-model build so polling always works; each model gets its own log.

run_bp_check:
- Add equals-style flag attempt (-metadata=, -module=, -model=, -packagesRoot=) as Attempt 2. xppbp.exe 10.0.24+ uses = separator; both previous attempts used : and silently fell through to help text, so BP check never actually ran on these environments.
- Element filter uses positional class:Name syntax for the equals-style attempt instead of the unsupported -filter: flag.
- Expand violation detection to catch BestPractices Warning:/Error: lines and non-zero Warnings:/Errors: summary lines — not just BPError prefix. Previously the status showed passed even when warnings were present.

mcpServer.ts (schema):
- Update build_d365fo_project schema: expose modelName, fullBuild, buildReferencedModels, force parameters and fix description. Previously only projectPath was visible to the model.

### DIFF
--- a/src/server/mcpServer.ts
+++ b/src/server/mcpServer.ts
@@ -1886,7 +1886,8 @@ SourceCode format for classes: class declaration with member vars inside { }, me
           type: 'object',
           properties: {
             projectPath: { type: 'string', description: 'Absolute path to the .rnrproj file to analyze. Auto-detected from .mcp.json if omitted.' },
-            targetFilter: { type: 'string', description: 'Optional: filter results to a specific class, table, or object name' },
+            targetFilter: { type: 'string', description: 'Optional: filter results to a specific object name (class, table, form, enum, ...).' },
+            targetElementType: { type: 'string', description: 'Element type for the filter when using xppbp 10.0.24+ (equals-style CLI). Common values: class, table, form, enum, view, query. Defaults to "class" when targetFilter is set but this is omitted.' },
             modelName: { type: 'string', description: 'Model name to check. Auto-detected from .mcp.json if omitted.' },
             packagePath: { type: 'string', description: 'PackagesLocalDirectory root path. Auto-detected from .mcp.json if omitted.' },
           },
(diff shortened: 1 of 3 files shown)


laeliand:
I leaved out full build on all deps but it could be added. The only time you need a full-tree full build is when a dep has stale symbols. When that happens, xppc reports it explicitly:

MetadataProvider Fatal Error: Class dynamics://Class/SomeClass has not been 
successfully compiled since it was last changed. Do a Full Build.
…and the queue stops on that dep. You then build that specific dep with fullBuild: true directly, then re-run the original call. It's one extra step but the error tells you exactly what to fix.

Adding a full-tree full-build parameter means paying the cost of full-compiling every dep in the chain every time — which in a deep tree could be many minutes — for a case that happens rarely and already has a clear remediation path.

So the four combinations you have now cover all practical cases well. The full-tree full-build is a "nice to have" edge case, not worth the added complexity and the runtime cost risk of accidentally triggering it?

PR #510, opened 2026-06-01
Fix: missing commands in setup doc and fixing build error

ahofNB:
Deployed on local dev box and had to make these changes to succeed.

### DIFF
--- a/bridge/D365MetadataBridge/NuGet.config
+++ b/bridge/D365MetadataBridge/NuGet.config
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="Microsoft Visual Studio Offline Packages" value="C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\" />
+  </packageSources>
+</configuration>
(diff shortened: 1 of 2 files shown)


dynamics365ninja:
Thanks for the contribution! The SETUP.md fix is clearly correct — `npm run build` is indeed missing from the setup steps and without it the server won't start. Happy to merge that part.

Before we merge the `NuGet.config` addition, could you help us understand a few things?

1. **What specific build error does this fix?** Could you share the exact error message you were seeing with `dotnet build`? The bridge project doesn't reference any NuGet packages — it uses D365FO DLLs via `HintPath`, so it's not immediately obvious what restore step was failing.

2. **The `<clear />` directive** will strip all globally configured package sources, which would break restore for anyone using private feeds (e.g. Azure Artifacts, Nexus). Was that intentional? If the only goal is to ensure `nuget.org` is present, we could use `<add>` without `<clear />`.

3. **The Visual Studio offline packages path** (`C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\`) is not present on all D365FO environments (cloud-hosted, UDE). Is this strictly required, or was it a leftover from local tooling?

If the root cause was a missing `nuget.org` source in a clean environment, a minimal fix without `<clear />` and without the VS offline path would be safer for all users. Happy to work through this with you!

ahofNB:
I retried the build in a new folder on same machine and now the original build succeeds. So, for now, I am not sure if something is different on the dev box now or if I made a mistake. So, its best to not submit that change.

PR #533, opened 2026-06-11
fix(security): prevent path traversal via workspacePath and filePath …

dynamics365ninja:
…params (GHSA-pw7v-9v7h-6j22)

- validateWorkspacePath: replace '..' substring check with assertReadRootAllowed root-containment check; absolute paths outside configured D365FO package roots (e.g. /etc, C:\Windows, ~/.ssh) are now rejected
- workspaceScanner.scanWorkspace: add per-file isFileUnderRoot re-check after glob to close the symlink escape vector
- formInfo filePath param: gate fs.readFile behind assertWritePathAllowed
- reportInfo filePath + sourcePath params: same gate, covers JSON indirection
- pathContainment: add assertReadRootAllowed() and isFileUnderRoot() exports

### DIFF
--- a/src/tools/formInfo.ts
+++ b/src/tools/formInfo.ts
@@ -14,6 +14,7 @@ import type { XppServerContext } from '../types/context.js';
 import { promises as fs } from 'fs';
 import { parseStringPromise } from 'xml2js';
 import { tryBridgeForm } from '../bridge/bridgeAdapter.js';
+import { assertWritePathAllowed } from '../utils/pathContainment.js';
 
 const GetFormInfoArgsSchema = z.object({
   formName: z.string().describe('Name of the form'),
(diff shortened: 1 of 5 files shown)


mcfly-zzh:
Hi @dynamics365ninja ,

Thank you for fixing this in PR #533.

I noticed that the fix branch and PR reference `GHSA-pw7v-9v7h-6j22`, and the PR notes indicate that the path traversal issue was addressed by adding root-containment checks for `workspacePath` / `filePath`, per-file checks after globbing, and guards for the `formInfo` and `reportInfo` file path parameters.

Would you consider publishing a GitHub Security Advisory for this issue, and requesting a CVE if you think it is appropriate?

Since the issue affected path handling in a published MCP server and has now been fixed, a published advisory would help downstream users identify the affected versions / commits and upgrade safely.

I am happy to help with further information if needed.

Thanks again for the fix.

PR #554, opened 2026-06-15
Regenerate runtime metadata after xppc build so new classes deploy without VS

laeliand:
Summary
An MCP-only xppc build produces the compiled .netmodule but does not update the binary .md runtime manifests the AOS uses to resolve class names — so newly added classes were invisible after deployment or deployable package until a full Visual Studio build was run manually.

xppc.exe compiles X++ to .netmodule but does not regenerate the .md manifests; in a normal build that step is performed by the Visual Studio MSBuild BuildTask, which isn't available in a headless/MCP build. Rather than reimplement manifest generation, this change invokes the same Microsoft API the BuildTask uses — RuntimeMetadataWriter.WriteAll in Microsoft.Dynamics.AX.Metadata.Storage.dll — directly, so the output matches what VS would produce.

Changes
generateMetadata.ts — after a successful compile, runs a cached .NET helper (RuntimeMetadataWriter.WriteAll) to rebuild the .md manifests from XML source.

xppc writes a compiled class's X++ compiler metadata to the -compilermetadata root, while the writer reads it relative to the -metadata (source) root; for a newly added class these differ, so the writer threw "Compiler metadata not set before serialization to runtime format." Fixed by overlaying the freshly written XppMetadata from the compiler-metadata root onto the source root before generation (overlay, not mirror — orphan metadata with no source is inert). 

The helper exe is cached under a name hashed from its C# source, so it rebuilds automatically when the helper source or framework version changes.

buildProject.ts — invokes runtime-metadata regeneration after the build queue completes; a regeneration failure is logged but does not fail the build.

runBpCheck.ts — points -compilerMetadata at the custom packages path so xppbp finds compiler metadata for newly added classes instead of reporting CompilerMetadataMissing.

Testing
Verified end-to-end on a live D365FO 10.0.2527 UDE environment, reproducing the original failure first and confirming the fix resolves it.

Original symptom

Added a brand-new X++ class (never built in Visual Studio) to a custom model.

Ran an MCP-only full build — xppc reported Errors: 0, Warnings: 0 — and deployed to the target environment.

The change was not actually delivered: the new class was absent from the running AOS despite the clean build and deploy.
Cause: the binary .md runtime manifests (which the AOS uses to resolve class names) were out of sync — xppc produces the .netmodule but never updates them. 

Added a component (generateMetadata.ts) to regenerate the .md files from XML source after each build, replicating the VS BuildTask post-compile step. (xppc alone does not do this, and the VS BuildTask that does is not usable headlessly — hence calling the underlying Microsoft API directly.)

Secondary failure surfaced by that component

With .md regeneration in place, RuntimeMetadataWriter.WriteAll then threw "Compiler metadata not set before serialization to runtime format" for the new class.

Root cause: xppc writes a new class's compiler metadata (XppMetadata\…\<Class>.xml) to the -compilermetadata root, while the writer reads it relative to the -metadata (source) root. The new class had source but no co-located compiler metadata → IsCompilerMetadataSet = false.

Fix: overlay the freshly written XppMetadata from the compiler-metadata root onto the source root before regeneration.

Fix verification (isolated)

Copied the missing XppMetadata entry into the source root and re-ran the helper → succeeded; the regenerated *_AxClass.md grew to include the new class. Diagnostics showed Already set: 48, Fixed: 0 — all classes carry compiler metadata.

Fix verification (real build path)

Removed the class, added a different new class with a unique marker, and ran a clean MCP-only full build through build_d365fo_project (patched server loaded).

Build reported ✅ Runtime metadata (.md) regenerated; the overlay sync ran automatically, the new class appeared in the regenerated *_AxClass.md, and the removed class was correctly absent.

Deployment confirmation

Deployed to UDE with no Visual Studio build and ran the new class — its marker output appeared in the AOS, confirming the class is fully resolvable at runtime from an MCP-only build.

Overlay-vs-mirror check

Confirmed orphan compiler metadata (an XppMetadata entry with no matching source class) is inert — the disk provider enumerates classes from source XML, so leftover entries never appear in the manifest. This is why the sync overlays rather than wiping the destination first.

Additional fix — runBpCheck.ts (related, independently useful)
BP check no longer emits CompilerMetadataMissing for newly-added classes once -compilerMetadata points at the custom packages path. Same underlying theme as the main fix (compiler metadata for new classes living under the custom path rather than the framework path), but it affects the BP-check tool rather than the build/metadata-regeneration path.

Scope / not covered
- This change addresses the runtime .md metadata manifests for AOT elements (the cause of new classes being invisible after an MCP-only build). Label compilation and SSRS report deployment are separate build pipelines that this change does not modify and were not verified here. Whether MCP-only builds correctly deliver new labels and new reports without a VS build is a known follow-up for me to test separately at a later time. 
- CHE was not used for testing
- Only Class objects were used in the testing, but should not be limited to.

### DIFF
--- a/src/tools/buildProject.ts
+++ b/src/tools/buildProject.ts
@@ -8,6 +8,7 @@ import crypto from 'crypto';
 import { getConfigManager } from '../utils/configManager.js';
 import { forceReleaseLock } from '../utils/operationLocks.js';
 import { lookupErrorFix } from './d365foErrorHelp.js';
+import { generateRuntimeMetadata } from './generateMetadata.js';
 
 const execFileAsync = util.promisify(execFile);
 
@@ -569,6 +570,26 @@ async function spawnXppcForState(ctx: XppcBuildContext, state: BuildJobState): P
         await writeBuildState(errState, customPackagesPath).catch(() => {});
       });
       return;
+    }
+
+    // All models built — regenerate .md runtime metadata manifests from XML source.
+    // xppc produces the compiled .netmodule but does NOT update the binary .md
+    // manifests that the AOS uses to resolve class names at runtime. Without this
+    // step, newly added classes are invisible to D365 after deployment even though
+    // the assembly compiled successfully.
+    const metaResult = await generateRuntimeMetadata(
+      microsoftPackagesPath,
+      customPackagesPath,
+      liveState.targetModel,
+    );
+    if (metaResult.skipped) {
+      await buildLog('WARN', `Runtime metadata regeneration skipped: ${metaResult.message}`);
+    } else if (metaResult.success) {
+      await buildLog('INFO', `Runtime metadata regenerated: ${metaResult.message}`);
+      await appendFile(state.logFile, `\n✅ Runtime metadata (.md) regenerated for ${liveState.targetModel}\n`, 'utf-8').catch(() => {});
+    } else {
+      await buildLog('WARN', `Runtime metadata regeneration failed (build still succeeded): ${metaResult.message}`);
+      await appendFile(state.logFile, `\n⚠️ Runtime metadata (.md) regeneration failed — VS build required for deployment of new classes:\n${metaResult.message}\n`, 'utf-8').catch(() => {});
     }
 
     // All models built — finalise as succeeded
--- a/src/tools/generateMetadata.ts
+++ b/src/tools/generateMetadata.ts
@@ -1,0 +1,275 @@
+/**
+ * generateMetadata.ts
+ *
+ * Regenerates D365FO runtime metadata manifests (.md files) from XML source
+ * after an xppc compile, without requiring a full VS build.
+ *
+ * Background: xppc.exe compiles X++ source into .netmodule files but does NOT
+ * update the binary .md manifests that the AOS uses to resolve class names at
+ * runtime. VS BuildTask does this as a post-compile step using the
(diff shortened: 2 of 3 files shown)


dynamics365ninja:
Thanks for tracking this one down, @laeliand — this is a genuinely nasty class of bug: a clean `xppc` build reporting `Errors: 0` while the change silently never reaches the running AOS. The kind of thing that costs hours precisely because everything *looks* green.

A few things I really appreciate here:

- **Calling the Microsoft API directly** (`RuntimeMetadataWriter.WriteAll`) instead of reimplementing manifest generation — matching exactly what the VS BuildTask produces is the right call and avoids us drifting from MS behaviour over framework versions.
- **The overlay-not-mirror reasoning.** The root-cause write-up (compiler metadata landing under `-compilermetadata` vs the writer reading from the `-metadata` source root) is clear, and the explicit check that orphan compiler metadata is inert — so overlaying is safe and wiping isn't needed — is exactly the kind of detail that prevents a "fix" from introducing a subtler bug later.
- **End-to-end verification on a live UDE**, reproducing the original failure first and confirming the new class actually resolves at runtime. The isolated + real-build-path + deployment checks leave very little doubt.

The `runBpCheck.ts` `-compilerMetadata` fix is a nice bonus — same underlying theme, independently useful.

Noted the scope caveats (labels/SSRS as a separate follow-up, Class-only testing) — those are reasonable boundaries to draw.

Issue #704, opened 2026-07-19
Incremental EXTRACT_MODE (custom/standard) rebuilds are slower than a full 'all' rebuild on a UDE single-language instance — two root causes


kklyandev:
## Summary

On a UDE environment, running `build-database` with `EXTRACT_MODE=custom` or `EXTRACT_MODE=standard` was **slower** than a full `EXTRACT_MODE=all` rebuild, not faster. Two independent issues combine to cause this:

1. **The build phase cannot see UDE auto-detected custom models**, so `custom` mode falls back to treating *every* model in the metadata directory as custom and rebuilds all of them.
2. **The incremental re-index runs one `indexMetadataDirectory` call per model in a loop**, whereas `all` does a single bulk pass — so even a correctly-scoped incremental build pays a large per-model overhead.

The result is that the documented "custom = fast path" does not hold in this configuration. Correctness is unaffected (all three modes produced an identical database); this is purely a performance/scoping problem.

## Environment measured

One of our UDE developer environments (name omitted). Single label language (`en-US`, `INCLUDE_LABELS=true`), so the labels database is small (~300 MB). Symbol database ~2.2 GB, 176 models, ~1.18M symbols. Of the 176 models, exactly 1 is custom.

## Measurements

Each mode was run as extract-metadata + build-database (the exact commands the rebuild flow uses), back to back, on the same machine. Database state was captured after each run.

| Mode | extract | build | **total** | DB after (models / symbols / size) |
|---|---|---|---|---|
| `custom` | 5.6 s | 772.9 s | **778.5 s (~13.0 min)** | 176 / 1,179,945 / 2201 MB |
| `standard` | 372.8 s | 754.1 s | **1126.9 s (~18.8 min)** | 176 / 1,179,945 / 2201 MB |
| `all` | 418.1 s | 220.0 s | **638.1 s (~10.6 min)** | 176 / 1,179,945 / 2201 MB |

What this means:

- **`all` was the fastest**, and **`standard` the slowest**. The intended fast path (`custom`) was ~22% slower than a full rebuild.
- **Extract scales as expected**: `custom` extract is 5.6 s vs `all` 418 s — the incremental extract is doing the right, small amount of work.
- **The inversion is entirely in the build/index phase**: `custom` and `standard` builds take ~750–770 s while the full build takes 220 s.
[...]

PR #711, opened 2026-07-20
Scope UDE custom extract by path so source ISV models are reindexed (follow-up to #704/#706)

kklyandev:
Follow-up to #704 (fixed by #706). #706 is a real improvement — the per-model reindex loop is gone (a `standard` build dropped from ~754s to ~140s for me) and the extract→build manifest bridge works well. But on UDE the manifest still under-classifies, so a `custom` refresh under-scopes. This PR closes that gap; the classification question behind it is laid out below for you to weigh.

## Problem

On a UDE environment the ModelStoreFolder (custom root) holds our own model plus third-party ISVs shipped **with X++ source** (e.g. Docentric AX and its satellite packages). After #706, a `custom` extract + build recorded only **one** model as custom:

    { "extractMode": "custom", "environment": "ude", "customModels": [ "MyModel" ] }
    Scoping custom build to 1 model(s) from extract manifest

Build-side scoping is correct — it faithfully reindexes what the manifest lists. The gap is upstream: `extract-metadata.ts` is only half path-based on UDE. The root scan (`~:471`) and the manifest `isCustom` flag (`~:550`) use the path rule (`rootPath === customRoot`), but the model-level filter (`~:538`/`:542`) still calls name-based `isCustomModel(modelName)` even when `customRoot` is set. Since `CUSTOM_MODELS` is empty on UDE and `isCustomModel()` matches only `D365FO_MODEL_NAME` (the `EXTENSION_PREFIX` branch is a case-sensitive `startsWith`), every ISV under the custom root is dropped as "standard" before it can reach the manifest.

Effect: a `custom` refresh after an ISV upgrade or pull reindexes only the configured model and leaves the rest of the source models under the custom root stale. (Binary-only ISVs are correctly out of scope in every mode — no AOT XML, so the `MODEL_MARKER_DIRS` gate skips them. This only concerns ISVs that ship source.)

## Fix

Compute the custom classification once using the same path rule already applied to the manifest flag, and use it for both the `custom-only`/`standard-only` guards and the manifest:

    const isCustom = customRoot ? rootPath === customRoot : isCustomModel(modelName);

On UDE, "custom" now means "lives under the custom root"; on traditional environments (no custom root) behaviour is unchanged. After the change, the manifest lists every source model under the custom root and the `custom` build scopes to them instead of just one.

## The classification question (for your call)

The fix rests on treating source ISV models as **custom**. That's a design decision, so here's the trade-off as I see it — happy to adjust if you weigh it differently.

**For (ISV = custom):**

1. **Refresh scoping** — the point of this PR. After pulling a new ISV build or a platform upgrade, a `custom` refresh should reindex ISV source; if ISV is "standard," the fast path silently skips those changes and the index goes stale.
2. **Pattern-catalog hygiene** — the form-pattern miner learns "standard D365 patterns" from *standard* models. ISV forms carry third-party conventions; classifying ISV custom keeps that corpus Microsoft-only.

**Considered but arguable both ways:** the custom-model boost in `contextRanker`. Boosting ISV helps surface it as part of the solution when answering questions — but your own model is arguably more relevant than a vendor's, so a single "custom" boost over-weights ISV relative to your own code. We don't lean on this either way; noting it so it's clear the spot was considered.

**Against (ISV = standard):** the write guard refuses edits to standard models to prevent modifying code you don't own; classifying ISV custom removes it from that guard.

**Why we weigh the write-guard concern lower:**

- The server can't reliably tell "our source" from "third-party ISV" — both are non-Microsoft siblings under the same custom root. "Microsoft vs non-Microsoft" is path-derivable (all that refresh + pattern-mining need); "ours vs ISV" is per-project developer knowledge that belongs in project config/instructions, not server-derived classification.
- An accidental ISV edit isn't silent — in a version-controlled project it shows up as a changed file under a vendor folder in the diff, and it's recoverable by redeploying the package.
- The high-risk case stays protected: Microsoft base models remain standard, so the write guard still hard-blocks edits to the base application.

If you'd rather keep ISV on the standard side for the write guard's sake, the alternative is to make the guard key on "is this the intended target of the write" rather than on `isStandardModel()` — but that's a larger change; this PR is the minimal one that fixes the refresh scoping.

### DIFF
--- a/scripts/extract-metadata.ts
+++ b/scripts/extract-metadata.ts
@@ -534,20 +534,27 @@ async function extractMetadata() {
         continue;
       }
 
+      // Custom classification for this model. On UDE (customRoot set) it is
+      // PATH-based — a model is custom iff it lives under the custom root —
+      // matching the root-level scan above and the manifest flag below. Only fall
+      // back to name-based isCustomModel() for traditional environments that have
+      // no custom root path. Using isCustomModel() here on UDE would drop ISV
+      // models with source under the custom root (their names match neither
+      // D365FO_MODEL_NAME nor EXTENSION_PREFIX), leaving a `custom` build scoped
+      // to just the configured model.
+      const isCustom = customRoot ? rootPath === customRoot : isCustomModel(modelName);
+
       // Apply model-level filtering
-      if (FILTER_MODE === 'custom-only' && !isCustomModel(modelName)) {
+      if (FILTER_MODE === 'custom-only' && !isCustom) {
         log.detail(`${glyph.arrow} skip standard model: ${modelName}`);
         continue;
       }
-      if (FILTER_MODE === 'standard-only' && isCustomModel(modelName)) {
+      if (FILTER_MODE === 'standard-only' && isCustom) {
         log.detail(`${glyph.arrow} skip custom model: ${modelName}`);
         continue;
       }
 
       const expectedXmlFiles = await countModelXmlFiles(modelDirsByLowerName);
-      // In UDE mode: custom iff the package lives under customRoot.
-      // In traditional mode: fall back to name-based detection.
-      const isCustom = customRoot ? rootPath === customRoot : isCustomModel(modelName);
       modelWorkItems.push({ packageName, modelName, modelPath, expectedXmlFiles, isCustom });
     }
   }
--- a/scripts/extract-metadata.ts
+++ b/scripts/extract-metadata.ts
@@ -45,6 +45,40 @@ const EXTRACT_MODE = process.env.EXTRACT_MODE || 'all';
 const isCustomModel = checkIsCustomModel;
 
 /**
+ * Decide how one model relates to "custom" for this extract run.
+ *
+ * `isCustom` — the classification recorded in the extract manifest and used for
+ * sourcePath normalisation. On UDE (customRoot set) it is PATH-based: a model is custom
+ * iff it lives under the custom root, matching the root-level package scan. Name-based
+ * isCustomModel() is only the fallback for traditional environments with no custom root.
+ * Using isCustomModel() on UDE would drop ISV models that ship source under the custom
+ * root — their names match neither D365FO_MODEL_NAME nor EXTENSION_PREFIX — leaving a
+ * `custom` build scoped to just the configured model (#711).
+ *
+ * `narrowedByConfig` — an explicit CUSTOM_MODELS list still NARROWS a custom-only run on
+ * UDE. The path rule decides what CAN be custom; a hand-maintained list decides how much
+ * of it to extract. The root-level scan ignores CUSTOM_MODELS entirely once customRoot is
+ * set, so without this an operator who set CUSTOM_MODELS="Contoso*" would have no way to
+ * scope a refresh to their own models and would re-extract every ISV under the root.
+ * Only wildcard patterns reach here — exact names take the MODELS_TO_EXTRACT path, which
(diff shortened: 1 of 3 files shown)


dynamics365ninja:
PR audit — the diagnosis and the fix both hold up; I verified them against the code. The root-level scan (`~:471`) does let ISV packages under the custom root through on UDE, and it was the model-level guard that dropped them by name, so the chain you describe is accurate. `rootPath === customRoot` is a safe comparison (both values come from `xppConfig.customPackagesPath`), the manifest flag is semantically unchanged, and `tsc --noEmit` is clean.

I pushed two commits to the branch:

**1. Regression: `CUSTOM_MODELS` wildcards stopped working entirely on UDE.** The root scan already ignored the patterns, but the model-level guard applied them — after the change there was nowhere left for them to take effect. `EXTRACT_MODE=custom` + `CUSTOM_MODELS="Contoso*"` would then extract everything under the custom root, and an operator would lose the only way to scope a refresh. 48784bf restores the patterns as a **narrowing** filter on top of the path rule: the path decides what *can* be custom, an explicit list decides how much of it gets extracted. Classification (and therefore the manifest) stays path-based, so `build-database` scoping is untouched when `CUSTOM_MODELS` is empty. For the record, exact names never reach that branch — they take the `MODELS_TO_EXTRACT` path, which leaves `FILTER_MODE` at `all`.

**2. Tests + a docblock.** 92f6f3a pulls the classification out of `extractMetadata()` into an exported `classifyCustom(rootPath, customRoot, modelName, customModels?)` and covers both rules — UDE/traditional × custom/standard, plus `Contoso*` vs `DocentricAXCore` under the same root. Mutation-checked: reverting the narrowing fails 1 test, reverting the classification to name-based fails 4. 48784bf also documents what `ExtractManifest.customModels` does and does not mean — "non-Microsoft under the custom root", not "models you own".

---

On the *classification question* section — that analysis reasons about effects this diff doesn't actually have, in both directions. The manifest has exactly one consumer, `build-database.ts:136-153`; runtime classification never touches it:

- **Write guard** (`src/tools/modifyD365File.ts:1078`) calls `isStandardModel()` at server runtime and does not read the manifest. So the "against" argument is moot — ISV stays behind the guard exactly as before, and this PR costs you no protection. The whole "why we weigh this lower" defence is answering a problem that isn't there.
- **`contextRanker` boost** (`src/workspace/contextRanker.ts:188`) — also name-based, unchanged. Noting that the spot was considered is fine; this diff simply doesn't reach it.
- **Pattern-catalog hygiene** ("for" argument #2) — `src/metadata/symbolIndex.ts:2595` gates on `isStandardModel()` and runs inside `build-database`, where `CUSTOM_MODELS` is empty on UDE. ISV tables therefore keep getting mined into `property_stats` as "standard". This PR does **not** deliver that benefit — unifying classification for the miner is separate work.

So the real payoff is argument #1 alone, refresh scoping — which is a good enough reason to merge on its own. I'll leave the description to you; just make sure the changelog doesn't inherit a promise the code doesn't keep.

Two more things worth adding to the description:

- A `custom` refresh on UDE now covers every source ISV, so it runs longer and `clearModels` casts a wider net. Expected, but anyone running Docentric and its satellites will feel it — and the newly-restored `CUSTOM_MODELS` narrowing is the answer for them.
- You correctly identified the case-sensitive `startsWith` in `isCustomModel()` (`modelClassifier.ts:392`) as a contributing cause. I'm not pulling that fix in here; filed as #721.

PR #714, opened 2026-07-21
chore(release): publish via trusted publishing, guard npm 12 script blocking

dynamics365ninja:
Follow-up to the [npm install-time security & GAT bypass2FA changelog](https://github.blog/changelog/2026-07-08-npm-install-time-security-and-gat-bypass2fa-deprecation/). Two independent problems from that post; `d365fo-mcp@1.0.0` is already published, so both are addressed before they can bite.

## 1 — Trusted publishing replaces NPM_TOKEN

Granular tokens with *bypass 2FA* are being retired in two phases: **early August 2026** they stop covering sensitive operations (token management, package access, maintainers, trusted-publishing config), and **around January 2027** they lose direct publishing entirely — reduced to staging a publish that a human approves with 2FA.

The `publish-npm` job now authenticates via **OIDC trusted publishing**, which stores no secret at all. This only became possible now that the package exists on the registry — the trusted publisher is configured in the package's own settings (already done on npmjs.com: org `dynamics365ninja`, repo `d365fo-mcp-server`, workflow `release.yml`).

- `NODE_AUTH_TOKEN` / `NPM_TOKEN` removed, along with the graceful-skip branch that existed only for the unconfigured-secret case.
- `--provenance` dropped: trusted publishing attests provenance automatically for a public repo publishing a public package.
- `id-token: write` retained — under OIDC it is the actual credential, not just a provenance enabler.
- Verified Node 24.x currently ships npm 11.16.0, comfortably over the documented floor (npm ≥ 11.5.1, Node ≥ 22.14.0); noted in a comment so nobody lowers the Node line and silently drops back to token auth.

**After merge:** the `NPM_TOKEN` secret and the npm token itself can be deleted — nothing reads them any more.

## 2 — npm 12 blocks install scripts (the silent one)

npm 12 blocks lifecycle scripts unless the **root** package.json pre-approves them by exact `name@version` in `allowScripts`. `better-sqlite3` is a native module, so this decides whether the server works at all.

Measured under npm 12 against the real dependency:

| install | native binding | exit code |
|---|---|---|
| with `allowScripts` | built | 0 |
| without | **missing** | **0 — warning only** |

The install *looks* clean and the failure only appears much later, when the server opens the index. The repo's existing pins currently match the lockfile exactly (`better-sqlite3@12.11.1`, `esbuild@0.28.1`), so users are fine today — the risk is drift, since the declared range is `^12.9.0` and dependabot is enabled.

- **`tests/packaging/allowScripts.test.ts`** — fails the PR in both directions: a non-optional package with an install script that no pin covers, and a pin matching nothing in the lockfile (the stale-after-bump case). Optional packages are exempt: `fsevents` is macOS-only and degrades to polling. Both failure modes were verified by simulating a bump; each prints the exact line to add.
- **`doctor`** — reports a missing binding by name with the `npm install-scripts approve` fix. It opens an in-memory database rather than merely importing the module: better-sqlite3 resolves its addon **lazily**, so a plain `import` succeeds even with the binding deleted — the first version of this check was a false negative, caught by testing it with the `.node` file renamed away.

## Testing

- Full suite: **132 files, 1784 passed**, 2 skipped (mode switches).
- New guard verified green, then red against a simulated dependency bump, then restored.
- `doctor` verified in both states (binding present → ok; `.node` renamed → fail with the fix line).
- Publish path itself cannot be exercised until the next `v*` tag — that release is the real proof, and the first thing to check is that provenance still appears on the npm page.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -79,18 +79,30 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   # Publishes the CLI/server package as `d365fo-mcp` so `npx d365fo-mcp` works.
-  # Requires the NPM_TOKEN repository secret (npm automation token); the job
-  # skips gracefully while the secret is not configured.
+  #
+  # Authenticates via npm trusted publishing (OIDC) — no NPM_TOKEN, no secret to
+  # rotate. npm matches the short-lived id-token against the trusted publisher
+  # configured on the package (org dynamics365ninja, repo d365fo-mcp-server,
+  # workflow release.yml), so renaming this file breaks publishing until the
+  # publisher entry is updated to match.
+  #
+  # Granular tokens with "bypass 2FA" are being retired instead: from early
+  # August 2026 they stop covering sensitive operations, and around January 2027
+  # they lose direct publishing entirely.
+  # https://github.blog/changelog/2026-07-08-npm-install-time-security-and-gat-bypass2fa-deprecation/
   publish-npm:
     name: Publish to npm
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      id-token: write   # npm --provenance
+      id-token: write   # required — this is the OIDC credential npm exchanges
     steps:
       - name: Checkout code
         uses: actions/checkout@v7
 
+      # Trusted publishing needs npm >= 11.5.1 and Node >= 22.14.0; Node 24.x
+      # currently ships npm 11.16, so the bundled npm is enough. Downgrading the
+      # Node line below 22.14 would silently fall back to token auth and fail.
       - name: Setup Node.js
         uses: actions/setup-node@v7
         with:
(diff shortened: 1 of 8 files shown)


dynamics365ninja:
Added the npm documentation to this PR (commit 0f9e425).

The package shipped with v1.0.0 but no document mentioned it existed — the `npx d365fo-mcp …` lines already in SETUP.md and CONFIGURATION.md are about managing an existing installation, written before the package was published.

- **npm version badge** in the README header.
- **README → "The npm package"**: states both halves plainly — the management commands are available as `npx d365fo-mcp <command>`, and inside a checkout npx resolves to the local build (verified: npm resolves the current project's own `bin`, so the existing doc references were not silently repointed at the registry by the publish). It also says what the package is *not*: `setup`/`update`/`index` are guarded to require the repository, so from a bare npx they refer the reader back to the installer instead of half-configuring a machine.
- **QUICK_START**: a short note answering the question the one-liner now invites — "why not just `npx d365fo-mcp setup`?"

The checkout-free path (a team member connecting to a deployed Azure server, where the only local artefact is `.mcp.json`) is named as upcoming rather than described, since `connect` does not exist yet — documenting it now would promise a command that fails.

PR #727, opened 2026-07-21
fix(cli): create fresh instances in the top-level layout so they are discoverable

kklyandev:
## Issue

`d365fo-mcp instance add` wrote the new instance config to `instances/<name>/config/d365fo-mcp.json`, but `listInstances()` only looks for `instances/<name>/d365fo-mcp.json` (or a legacy `.env`). The freshly created instance was therefore invisible to `instance list`, `rebuild`, `run` and `upgrade` — the wizard produced an instance its own management commands could not see, and its "Next steps" pointed at commands that then reported the instance as not found.

Root cause: for a not-yet-created config, `resolveConfigFiles()` always fell back to the repo layout (`<baseDir>/config/d365fo-mcp.json`). Instances need the top-level instance layout that the docs, `instances/d365fo-mcp.template.json` and `add-instance.ps1` all promise.

## Fix

- **`resolveConfigFiles`**: add an optional `fallbackConfigPath` used only when neither candidate exists; the default is unchanged, so the repo root still lands on `config/d365fo-mcp.json`.
- **`openInstanceStore`**: pass the instance's own top-level `d365fo-mcp.json` as that fallback, so a fresh instance is written where discovery looks.
- **`listInstances` / `createInstance`**: recognise either candidate location, so an instance an older build already wrote under `config/` is still found (and the create-time guard still refuses to clobber it).

### DIFF
--- a/src/cli/instances.ts
+++ b/src/cli/instances.ts
@@ -5,6 +5,7 @@
  */
 import * as fs from 'node:fs';
 import { join } from 'node:path';
+import { configCandidates } from '../config/configFile.js';
 import { settingByPath } from '../config/settings.js';
 import { paths } from './context.js';
 import { openInstanceStore, readSetting, saveStore, writeSetting } from './settingsStore.js';
@@ -26,7 +27,10 @@ export function listInstances(): Instance[] {
     .filter(e => e.isDirectory())
     .map(e => {
       const dir = join(paths.instancesDir, e.name);
-      return { name: e.name, dir, envFile: join(dir, '.env'), configFile: join(dir, 'd365fo-mcp.json') };
+      // Accept either instance layout: the top-level d365fo-mcp.json, or the
+      // config/ form an older build's wizard wrote. Prefer the top-level one.
+      const configFile = configCandidates(dir).find(p => fs.existsSync(p)) ?? join(dir, 'd365fo-mcp.json');
+      return { name: e.name, dir, envFile: join(dir, '.env'), configFile };
     })
     .filter(i => fs.existsSync(i.configFile) || fs.existsSync(i.envFile))
     .map(i => {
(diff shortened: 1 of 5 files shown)


dynamics365ninja:
Audited this — the diagnosis holds and the fix is on the right layer. Verified locally: the premise (`resolveConfigFiles` hard-coded the repo fallback, so a fresh instance landed in `config/` while `listInstances` looked at the top level), a clean merge with current `main` (b17d778 touches `configFile.ts` but does not overlap), `tsc --noEmit` clean, and reverting `src/` makes two of the three new tests fail.

Every consumer of the config path — `target.ts` (`targetEnv`, `instanceTarget`), `doctor.ts`, `mcpJson.ts`, `commands/instance.ts` — goes through `store.configPath` / `openInstanceStore`, so both layouts behave the same, and `configBaseDir()` keeps `./data/…` anchored to the instance folder for the old one. Nothing left hard-coding the instance path.

One real gap, now pushed as c0a8760 rather than left as a comment:

**The `instances.ts` half had no test.** `still finds an existing instance config written under config/` passes on `main` too — it exercises `openInstanceStore`, which already probed both candidates; the new candidate preference in `listInstances()` and the widened guard in `createInstance()` were uncovered, and `src/cli/instances.ts` had no test file at all. `tests/cli/instances.test.ts` mocks the data root to a temp directory (the suite runs from a checkout, where `paths.instancesDir` is the repo's real `instances/`) and covers both layouts, the top-level preference when both exist, the legacy `.env`-only instance, and the create-time guard on either layout. Reverting `src/cli/instances.ts` fails two of the eight.

Two smaller things in the same commit:

- `createInstance` returned `join(dir, 'd365fo-mcp.json')` as `configFile` while `saveStore` wrote to `store.configPath` — the same duplicated knowledge this PR centralises. Now returns `store.configPath`.
- The new `fallbackConfigPath` comment says "used only when neither candidate exists"; it is also skipped when `D365FO_CONFIG` names the file outright. Harmless for the CLI (`allowEnvOverride: false` everywhere) but a trap for the next caller, so the comment now says so.

Not blocking, filed as #744: an instance already written under `config/` is now discoverable but has no way out of that layout, while `docs/SETUP.md`, `d365fo-mcp.template.json` and `add-instance.ps1` all tell the user its config is at the top level.

LGTM with those in. 531 tests pass in `tests/cli` + `tests/utils` on the merged tree.

PR #786, opened 2026-07-28
eval: class-inheritance case + close its coverage orphan

dynamics365ninja:
## What

`class-inheritance` was the only entry in `orphans.knowledge`. The topic landed in the KNOWLEDGE_BASE in #781, but nothing in the coverage taxonomy claimed it — so no leaf could ever go green on it, and the knowledge entry was invisible to the metric.

Two parts:

### 1. `eval/cases/L2-coc-inherited-method.json`

A CoC wrapper for a method the augmented class only **inherits**. This is precisely the shape that was broken this week: the tool path walked **zero** levels of the extends chain (#780/#781/#782), so an inherited method read as non-existent and the agent concluded a wrap was impossible.

Facts pinned by the case, verified against the VM:

```
# Method: SalesFormLetter.promptAndRun
Model: Foundation   Returns: void   Modifiers: public
ℹ️ Inherited method. SalesFormLetter_Invoice does not declare promptAndRun —
   it inherits it from SalesFormLetter.
```

Design points worth flagging for review:

- **The oracle is the doc comment.** The wrapper must name the declaring class and the full extends chain in its `<remarks>`. A golden that matches proves the agent *walked* the chain; it cannot be produced by guessing the method exists.
- **`[ExtensionOf]` deliberately targets the SUBCLASS.** Verified empirically against xppc: both targets compile, and a signature mismatch is reported against the *declaring* class. The narrow scope is the one under test, so the instruction forbids retargeting at the base.
- **No SysTest.** `SysTestConsole.exe` / `SysTestRunner.exe` do not exist on the eval VM, so runtime firing is out of scope; the oracle is the golden AxClass XML plus a clean build.
- The known `ModuleReferences`/ApplicationSuite Descriptor prerequisite is called out in the instruction, since it otherwise surfaces as a "does not specify a known class" build failure that has nothing to do with the case.

### 2. A matching taxonomy leaf

The code-side mirror of the existing `table-inheritance` leaf.

## Coverage impact

Core drops **100% → 97.7% (43/44)**. That is the taxonomy behaving as its header comment documents — a new leaf without K/E/T visibly drops the number rather than going unnoticed. `E` flips once the golden is captured on the VM, and core returns to 100%.

`orphans.knowledge` is now empty. The two entries in `orphans.cases` pre-date this branch.

## Verification

- `npm run eval:coverage -- --check` — ✅ clean
- `npx vitest run tests/eval/` — 338 passed
- `npx tsc --noEmit` — clean

## Next step

Capture the golden on the VM via the eval-implementer role (`golden_pending: true` until then).

https://claude.ai/code/session_0182Dqj7bduTYw1VSY8a8F4b


dynamics365ninja:
Pushed a second commit — the case as originally drafted had a **void oracle**, so it should not have been merged as-is.

It pinned the declaring class and the extends chain inside a `///` `<remarks>` block. But `canonicalizeXppDocComments` collapses every contiguous run of `///` lines to a single `/// <xmldoc/>` placeholder (deliberately — doc wording isn't pinned by any case, and `d365fo_file` auto-injects generated prose; oracle-fragility finding #24).

Measured against the real normalizer on the case's own wrapper shape:

```
golden CANNOT distinguish right from wrong declaring class? true

AxClass/SourceCode/Methods/Method[promptAndRun]/Source = /// <xmldoc/>
public void promptAndRun()
{
    next promptAndRun();
}
```

The case would have scored `golden_match: 1` for a declaring class of `TotallyWrongClass` — green, proving nothing. That is the very failure mode the case exists to catch, reintroduced inside its own oracle.

**Fix:** the two facts move into `//` line comments in the method body, which normalization leaves alone. The `///` `<summary>` header stays, since its *presence* is what `BPXmlDocNoDocumentationComments` measures. Verified:

```
declaring class survives normalization?     true
golden DISTINGUISHES wrong declaring class? true
golden DISTINGUISHES wrong chain?           true
```

`tests/eval/inheritedMethodCaseOracle.test.ts` asserts the property that matters (the oracle separates right from wrong) rather than the instruction wording, so it survives a reword — and it also pins the `///` collapse itself, so a change there forces the case's rationale to be revisited instead of silently rotting.

342 eval tests pass.

dynamics365ninja:
## Golden captured on the VM — core back to **100%**

`golden_pending: false`. **core 43/44 (97.7%) → 44/44 (100%)**, total 66/78 → **67/78 (85.9%)**. The leaf is now K✅ E✅ T✅.

Ground truth resolved from the symbol index, not assumed:

```
chain     SalesFormLetter_Invoice -> SalesFormLetter -> FormLetterServiceController
          -> SysOperationServiceController -> SysOperationController
declares  SalesFormLetter   (public void promptAndRun(), model Foundation)
```

### Both negative tests run

A clean build proves nothing on its own here, so:

**1. Build.** Poisoning the wrapper to `promptAndRun(int)` fails:

> `The augmented class 'SalesFormLetter' provides a method by this name, but this method cannot be used as a chain of command method since the parameter profile does not match the original method.`

The compiler names the **declaring** class even though `[ExtensionOf]` names the **subclass** — the case's central claim, re-confirmed live rather than recalled from earlier in the session.

**2. Oracle.** `golden_match` against the real captured golden:

| mutation | golden_match | changed paths |
|---|---|---|
| identical (control) | **1** | 0 |
| wrong declaring class | 0 | 1 |
| truncated chain | 0 | 1 |
| `[ExtensionOf]` retargeted at base | 0 | 1 |
| facts deleted entirely | 0 | 1 |

The golden is load-bearing. The pre-fix `///` version would have scored **1 on all four**.

### Two tool defects found by running the case

Both recorded in `eval/corpus/runs/2026-07-28T15__L2-coc-inherited-method__a8a6204.json`. Neither is fixed here — flagging for triage.

**`build_d365fo_project` replays a stale result as a fresh success.** After a wait-timeout handoff, a second call returned `✅ Build succeeded / Errors: 0` with *byte-identical* phase timings from the previous build (`Metadata Validation | 34,754`, `Compilation ended | 122,906`, `00:02:02`). Proof it never ran:

```
poisoned XML written   15:05:46
build log last written 15:05:04   <- 42 s EARLIER
```

`force: true` produced the real answer (2 errors). This is invisible without checking log mtimes — **any `pass@build` not taken with `force: true` is weaker than it looks.**

**`prepare(mode="change")` is still inheritance-blind.** For `SalesFormLetter_Invoice.promptAndRun` it reports `(not found in symbol index)` and `CoC eligibility: (could not determine — method not found in symbol index)`. `get_method` walks the chain correctly after #780/#781/#782; `prepare` — the one-call aggregator agents are told to *start* from — still does not. Same defect class as the original bug, on an unfixed surface.

### Sandbox

Rolled back: file deleted, project entry removed, index cleaned. A grep across `K:\repos` and `K:\AosService` finds zero residual references, and a final full build of Contoso is green (`Errors: 0`).

343 eval tests pass · `eval:coverage --check` clean.

dynamics365ninja:
## Fixed the `prepare` defect found by this case

`prepare(mode="change")` now walks the extends chain. This was the second defect the eval run surfaced, and it was the more important of the two: `prepare` is the aggregator agents are instructed to **start** from, so the original inheritance bug survived exactly where it does the most damage — an agent reads "the method does not exist" and abandons the CoC path before ever reaching the tools #780/#781/#782 fixed.

**Before** (`SalesFormLetter_Invoice.promptAndRun`, live on the VM):

```
### Method signature _(symbol index)_
(not found in symbol index)

### CoC eligibility
(could not determine — method not found in symbol index)
```

…while `get_method` on the same pair correctly reported it as inherited from `SalesFormLetter`.

### Change

Both call sites (`fetchMethodSignature`, `fetchEligibility`) shared the same `parent_name = <objectName>` probe — declared members only. They now share one `readMethodRow` that retries against the declaring ancestor via the existing `findDeclaringAncestor`. Index-safe: `parent_name` stays BINARY on `idx_parent_type_name`; `COLLATE NOCASE` applies only within one object's member range.

Eligibility for an inherited method **names both wrap targets** rather than picking one, because xppc accepts either — verified against the compiler earlier in this work. They differ only in scope, and a signature mismatch is reported against the *declaring* class whichever is named, so that is stated too:

```
✅ Method appears CoC-eligible — inherited from `P_Base`.
Pick the target deliberately; both compile:
- `[ExtensionOf(classStr(P_Leaf))]` — wraps it for `P_Leaf` only.
- `[ExtensionOf(classStr(P_Base))]` — wraps it at the declaration, so it runs for **every** subclass of `P_Base`.
Either way the signature must match `P_Base`'s declaration exactly …
```

### Verification

The test chain is two levels deep on purpose — a fix consulting only the direct parent still fails the grandparent case. **Confirmed red before the fix:**

```
× finds a method declared on the DIRECT parent
× finds a method declared on the GRANDPARENT (a direct-parent-only fix fails here)
× reports CoC eligibility for an inherited method instead of "could not determine"
× names both wrap targets, since xppc accepts either
  Tests  4 failed | 2 passed (6)
```

The two that stay green in both states are the controls: a declared method must not be labelled inherited, and a genuinely absent method must still report not-found.

**1524 tests pass** (101 files) · `tsc --noEmit` clean · `dist/` rebuilt.

> Note: the running MCP server serves `dist/`, so this needs a **reconnect** before the live behaviour changes.

The other defect from the run — `build_d365fo_project` replaying a stale result as a fresh success — is **not** fixed here. It is recorded in the corpus record and worth its own PR; it silently weakens every `pass@build` not taken with `force: true`.

dynamics365ninja:
## Fixed the stale-build defect too

Both defects this eval run surfaced are now fixed on this branch.

### The bug

A **finished** build state left on disk was returned verbatim to the *next* call, which then read as that call's own result. Nothing was compiled. The existing `buildModeChanged` guard didn't help — it only covers incremental → full, and both runs here were full.

The tell was the timings:

```
✅ Build succeeded (xppc.exe, full build (target))
Metadata Validation | 34,754 | 2022
Compilation ended   | 122,906 | 2023
Errors: 0
```

byte-identical to the previous run. Two real builds never share millisecond timings. Proof nothing ran:

```
poisoned XML written   15:05:46
build log last written 15:05:04   <- 42 s EARLIER
```

Only `force: true` produced the truth (2 errors).

### The fix

The caller's intent is unknowable — *"collect the build I started"* and *"build now"* are the same call, so no flag can separate them. The decision is keyed on the **disk** instead: a finished result may be reused only when no source changed after it ended. Otherwise it is discarded and a real build runs. When it *is* reused, the output now says so with the end time, so a collected result can never be read as a fresh one:

```
ℹ️  Collected the result of the build that ended 2026-07-28T15:05:04.915Z
    (no source changes since — nothing was recompiled by this call).
```

Details worth reviewing:

- **`bin/` and `XppMetadata/` are excluded.** They are build *outputs*, always newer than the build that wrote them — scanning them would mark every result stale and rebuild forever. Verified against the real Contoso package: a full scan right after a build finds no change.
- **Directory mtimes are checked**, which is what catches a **deletion**. A removed class leaves no file to stat and would otherwise read as "nothing changed" — the same stale-green failure from the other direction.
- **Unreadable tree or blown time budget ⇒ "changed".** The two error directions are not equal: a needless rebuild costs minutes; a wrongly reused result reports a compile that never happened.
- **Cost on the real tree: 48 ms** worst case (full scan, nothing newer) against a 2-minute build.

### Verification

`tests/tools/buildStaleResult.test.ts` — 10 cases covering the incident timeline, nested changes, the output-dir exclusion, deletion, and every refusal path.

`buildProject.test.ts` needed `stat()` in its `fs/promises` mock plus re-arming after `resetAllMocks()` — an unarmed `stat` resolves `undefined`, which the scan correctly reads as "unreadable, assume changed".

**178 test files, 2356 tests pass** · `tsc --noEmit` clean · `dist/` rebuilt.

> Needs a **reconnect** to take effect live — the server serves `dist/`.

Worth noting for triage: this defect silently weakened every `pass@build` in the corpus that was not taken with `force: true`. Past green runs aren't necessarily wrong, but they're weaker evidence than they look.

dynamics365ninja:
## Goldens re-verified against the compiler — **65/65 clean**

The stale-replay defect meant any `pass@build` not taken with `force: true` could have described a tree nobody compiled. Golden *content* was never at risk (it is captured from disk), but the claim that the captured artifacts **compile** rested on exactly the signal the defect could fake. That claim is now re-established against xppc itself.

```
65/65 clean · 0 with errors · 143 artifacts
```

**No golden was invalidated.** Evidence: `docs/golden-build-reverification-2026-07-28.md`, per-case data in `eval/golden-build-verification.json`, repeatable via `scripts/verify-goldens-build.ts`.

### Method

Per case, **in isolation**: write that case's goldens into the Contoso sandbox → full `xppc` build (no `-incremental`, so metadata validation actually runs) → collect errors → delete what was written. xppc is invoked directly, without the state file whose replay bug started this. The baseline sandbox with nothing written is clean, so every error is attributable.

### Isolation is load-bearing — the bulk number is a trap

I tried a bulk pass first (all 143 artifacts at once). It reported **12 errors, of which the isolated runs showed 0 were real**:

| Bulk error | Actually |
|---|---|
| `AxEdtString/ConDemoNoteSubject: Name conflict; 'AxClass' is using the same name` | `L0-edt-basic` and `L2-delegate-basic` legitimately reuse the name for different object types. They never coexist in a real run. |
| `ConNumberSeqModuleDemoNote: qualifier 'NumberSeqModule' is not valid for field 'ConDemoNote'` | `L2-numberseq-basic` needs **its own** `NumberSeqModule.ConExtension`; the bulk pass kept the empty-values one from another case. |
| `Field group 'Overview' does not exist` ×3 | The bulk pass **overwrote the `ConDemoNoteHeader` fixture** with a case's golden copy of it, breaking three unrelated forms. |
| Workspace pattern errors ×4, `ConDemoNoteMap/CreatedDateTime`, `ConDemoNoteHeaderLine/LineNum` | Same contamination. Clean in isolation. |

Bulk-clean would also have been the *weaker* claim, since one case's object can satisfy another's dangling reference. Both directions are wrong, so the bulk number is not reported as a result anywhere. The script now never overwrites an existing file, for the same reason.

### Stated, not quietly counted as green

Two artifacts were **not** compile-verified: `AxTable/ConDemoNoteHeader` in `L1-form-basic` and `L2-dimension-basic`. It is the VM fixture, so it was left alone and those cases compiled against it — which is what a real run does. Their forms and classes are verified; those two golden tables are not.

### Environment change left in place

`FleetManagement` added to the sandbox Descriptor's `ModuleReferences`. `L2-coc-extension` wraps `FMVehicleDataContract` and xppc does not resolve package references transitively for directly-named types, so the case cannot build without it — the same fix already applied for ApplicationSuite, Directory, Ledger, ContactPerson and Currency. Revert if unwanted.

### Sandbox

Clean. Every written artifact removed after its build; the fixture is untouched (original mtime `2026-07-21 18:10:59` intact); no manifest or backup files remain.

PR #799, opened 2026-07-31
fix(bridge): add table-extension fallback for relation/index write ops

laeliand:
AddRelation, RemoveRelation, AddIndex, and RemoveIndex only ever called _provider.Tables.Read(tableName), so any of these ops targeting a table-extension threw "Table not found" inside the bridge process. The exception was then swallowed by a generic catch in bridgeAdapter.ts, surfacing as a misleading "returned null - required parameters may be missing" even when every required param was present.

Mirror the existing Tables -> TableExtensions fallback already used by AddField, adding it to these four ops so they work on both plain tables and table-extensions.

### DIFF
--- a/bridge/D365MetadataBridge/Services/MetadataWriteService.cs
+++ b/bridge/D365MetadataBridge/Services/MetadataWriteService.cs
@@ -1832,13 +1832,9 @@ namespace D365MetadataBridge.Services
         // TABLE INDEX OPERATIONS
         // ========================
 
-        /// <summary>Adds an index to a table.</summary>
+        /// <summary>Adds an index to a table or table-extension.</summary>
         public object AddIndex(string tableName, string indexName, List<string>? fields, bool allowDuplicates, bool alternateKey)
         {
-            var axTable = _provider.Tables.Read(tableName)
-                ?? throw new ArgumentException($"Table '{tableName}' not found");
-            var msi = GetModelSaveInfoForObject(_provider.Tables, tableName);
-
             var axIdx = new AxTableIndex { Name = indexName };
             axIdx.AllowDuplicates = allowDuplicates ? Microsoft.Dynamics.AX.Metadata.Core.MetaModel.NoYes.Yes : Microsoft.Dynamics.AX.Metadata.Core.MetaModel.NoYes.No;
             if (alternateKey)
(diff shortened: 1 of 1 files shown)


dynamics365ninja (commented):
## Audit — looks right, and it compiles against the real metamodel

Since `eval-gate` is Node/TS only and **no CI job compiles the bridge**, I built this branch in an isolated worktree against the metadata assemblies on the dev VM rather than reading the diff alone:

```
Bridge compiles against metamodel 7.0.7858.27 from K:\AosService\PackagesLocalDirectory\bin
Build succeeded.  0 Warning(s)  0 Error(s)
```

That confirms the central claim of `ed03086`: `axExt.Indexes`, `axExt.Relations` and `IMetaTableExtensionProvider` are all statically available, so the `dynamic` casts really were buying nothing but a loss of compile-time checking.

### What I checked

- **The fallback matches the one it mirrors.** All four extension branches are shape-identical to `AddField` (`MetadataWriteService.cs:1458-1481`), including the `as … ?? throw` provider cast — same as `ModifyField` / `RenameField` / `RemoveField` / `ReplaceAllFields`.
- **Static collection access is precedented, not assumed.** `CreateTableExtension` already does `axExt.Indexes.Add` / `axExt.Relations.Add` (`:957`, `:979`), and `RenameField`'s extension branch iterates `axExt.Indexes`. The captured golden `eval/goldens/L2-table-extension/CustGroup.ConExtension.metadata.xml` shows `<Indexes>` and `<Relations>` on a real `AxTableExtension`, so the serialiser handles them.
- **`Tables` before `TableExtensions` is unambiguous.** An extension name always contains a dot (`Base.ModelExtension`), a table name never can — the two namespaces cannot collide, so the ordering has no shadowing risk.
- **The fix is reachable from the MCP path, not dead code.** `table-extension` ∈ `BRIDGE_MODIFY_TYPES` and the four ops ∈ `BRIDGE_MODIFY_OPS` (`bridgeAdapter.ts:1189-1216`), so `canBridgeModify` lets these through.
- **`propertiesWritten` still works.** The new `AddRelation` extension branch echoes `cardinality` / `relatedTableCardinality` / `relationshipType`, which is what `bridgeAdapter.ts:1519` keys on — so `directXmlEnsureRelationProperties` correctly stays a no-op instead of patching the XML a second time.
- **The reworded error survives classification.** `"Table or table-extension 'X' not found"` still matches `isUnresolvedObjectError` (`modifyD365File.ts:895-900`) as a genuine resolution failure, so refresh+retry behaves as before.

### Two things left open (not blockers, not this PR's job)

1. **No test or eval case exercises the new path.** `L2-table-extension` covers field + field group on an extension via the **create** path only; nothing drives `add-index` / `add-relation` against a table extension. A green build says nothing about the written XML.
2. **The "created this session" hole stays open for extensions.** `directXmlAddIndex` guards on `/<AxTable\b/` (`modifyD365File.ts:660`), which by design does not match `<AxTableExtension`. So when the provider's startup-fixed metadata roots don't yet know a same-session extension — the exact scenario that function was written for (`:627-636`) — there is still no fallback. The extension XML shape is identical, so widening the guard is the same patch.

### Follow-up filed

The same defect — a write op resolving its target through a single provider collection — is still present in six more ops, each advertised as supporting the extension type it cannot serve: the three field-group ops (`Tables.Read` only), the three enum-value ops (`Enums.Read` only), and `AddDataSource` (`case "form"` only, while the schema documents form-extension support). The enum one is the sharpest: adding a value to a standard extensible enum *through an enum extension* is the only supported way to do it in D365FO, and the repair path can currently only throw.

Details, line references and a proposed fix in #803.

PR #800, opened 2026-07-31
Feat/form extension and data entity extension support

laeliand:
## Summary
- **Form extensions:** `add-data-source` had no `"form-extension"` case at all in the bridge (only `"form"`, via `_provider.Forms`), so it threw for every call against an extension. `add-control` already had a working direct-XML fallback for this same class of gap — added the equivalent fallback for `add-data-source`, writing an `<AxFormDataSource>` element straight into the extension's `<DataSources>` collection, in the field order confirmed against a real Microsoft-shipped `AxForm` data source.
- **Data-entity extensions:** `add-field` had zero support for `data-entity-extension` — not in `BRIDGE_MODIFY_TYPES` at all, so it was rejected before even reaching the operation dispatch. A mapped field (`AxDataEntityViewMappedField`) has no EDT of its own unlike a table field, so it needs its own param pair (`dataField`/`dataSource`) instead of `fieldType`/`fieldBaseType` — added as a new direct-XML fallback (the bridge has no provider path for this object type at all), plus the corresponding `required`/`optional`/`mutationOneOf` rules in the op-spec registry.
- **Guidance fix:** while chasing the above, found that both the `params` property description and the op-spec error text implied `params`-nested and flat-top-level calling styles are equally reliable ("pass inside `params` or flat at top level"). They are not — a caller whose MCP client validates strictly against the base wire schema will have any undeclared flat top-level key silently dropped before it reaches this server, then see a misleading "required parameters missing" error that has nothing to do with the real cause. Reworded both to make clear `params` is the only nesting new callers should rely on. Bumped the per-tool schema budget (documented, per that test file's own convention) to cover the added guidance text and the two new param specs.

## Test plan
- [x] `add-data-source` against a real `AxFormExtension` — confirmed correct `<AxFormDataSource>` XML (Name/Table/Fields/ReferencedDataSources/JoinSource/LinkType/DataSourceLinks/DerivedDataSources) written via the new fallback.
- [x] `add-control` against the same extension — confirmed still works via the pre-existing fallback (unaffected by this change).
- [x] `add-field` against a real `AxDataEntityViewExtension` with `{fieldName, dataField, dataSource}` — confirmed correct `<AxDataEntityViewMappedField>` XML written via the new fallback.
- [x] `npx vitest run tests/utils/toolSchemaBudget.test.ts tests/tools/d365foFileOpSpecs.test.ts` — both pass (schema stays under the (raised) per-tool cap; op-spec coverage intact). The pre-existing `13 vs 26` tool-count failure in that suite is a local 
- [x] `npm run build` — clean, no TS errors.


dynamics365ninja:
## Audit + rework pushed to this branch

Reviewed this against `main` on the dev VM and pushed the rework as a merge commit (4547ac4) rather than asking for another round — the branch had gone stale under #804 and two of the findings were silent-write bugs, which are the kind that pass a green build. Everything below was reproduced before it was changed. The diagnosis in the original PR was right in both halves; it is the implementation of the data-entity half that had to move.

### The form-extension half is gone — `main` got there first

#804 (ec07ca3) implemented `add-data-source` for `form-extension` **in the C# bridge**: `AxFormExtension.DataSources` holds the very same `AxFormDataSourceRoot` as `AxForm.DataSources`, so `AddDataSource` gained a real extension branch and reuses `CreateFormDataSourceRoot` unchanged (`MetadataWriteService.cs:3126`). Keeping a second, string-replacing copy in TS would only shadow the API path.

Worth saying explicitly: **the element order this PR established was correct.** `Name, Table, Fields, ReferencedDataSources, JoinSource, LinkType, DataSourceLinks, DerivedDataSources` — checked against shipped `ApplicationSuite\Foundation\AxFormExtension\BusinessProcessActionLookup.Foundation.xml:954-957`, where `JoinSource`/`LinkType` sit exactly there. That is what the bridge branch writes too.

### The data-entity half was reachable via the bridge after all

The PR's premise was that "the bridge has no provider path for this object type at all". Reflecting over the VM's metamodel says otherwise:

```
IMetadataProvider.DataEntityViewExtensions : IMetaDataEntityViewExtensionProvider
AxDataEntityViewExtension.Fields           : KeyedObjectCollection<AxDataEntityViewField>
AxDataEntityViewMappedField : AxDataEntityViewField
    { Name, DataField, DataSource, Label, Mandatory, ... }
IMetaDataEntityViewExtensionProvider : ISingleKeyedMetadataProvider<AxDataEntityViewExtension>
```

That last line is the nice part — `Read`/`Update` need no cast here, unlike Forms and Tables where they are explicit interface members. So `add-field` now goes through the bridge, same shape as everything #804 added, and the bridge routes on the `dataField`/`dataSource` pair rather than the object name. Direct XML stays only as the same-session escape hatch `add-index` and `add-control` already have.

### Three silent defects in the direct-XML writer

Each of these wrote something unusable and answered `✅`.

**1. Wrong `<Fields>`.** `content.replace('</Fields>', …)` takes the *first* match, and an `AxDataEntityViewExtension` carries `<FieldGroupExtensions>` — whose `AxTableFieldGroupExtension` has a nested `<Fields>` of its own — *before* the top-level one. Run against the shipped `CurrencyEntity.Extension.xml`:

```xml
<FieldGroupExtensions>
  <AxTableFieldGroupExtension>
    <Name>AutoReport</Name>
    <Fields>
      <AxTableFieldGroupField>…</AxTableFieldGroupField>
        <AxDataEntityViewField xmlns="" i:type="AxDataEntityViewMappedField">   ← landed here
```

This hits the *typical* case — an extension that already has one field — which is the main reason `add-field` exists. Replaced with a depth-counting `findTopLevelCollection` that only matches a direct child of the root.

**2. Sub-element order.** The PR emitted `Name → DataField → DataSource → Label`. Shipped elements put the presentation properties first:

```xml
<Name>DecimalsCount_MX</Name>
<CountryRegionCodes>MX</CountryRegionCodes>
<Label>@MexicoCFDI:DecimalsNumber</Label>
<DataField>DecimalsCount_MX</DataField>
<DataSource>Currency</DataSource>
```

`src/tools/dataEntityViewExtensionXml.ts:21` already documents why that matters — *"the metadata deserializer silently drops children it meets out of order"* — so `fieldLabel` was being lost. The element now comes from that same builder (`buildAxDataEntityViewFieldXml`, extracted from the create path) so the two cannot drift, which is the whole reason that file exists.

**3. Idempotency scope.** `new RegExp('<Name>' + fieldName + '</Name>')` was tested against the whole file. `<Name>` also carries the extension's own name, every field-group name and every `AxPropertyModification` — a false hit answered "already present" and wrote nothing. Now scoped to `AxDataEntityViewField` elements. (The `<Table>{table}</Table>` probe in the form-extension writer had the same shape; it left with that writer.)

### Also changed

- **`fieldGroupName`** (optional) appends the field to a base-entity group via `<FieldGroupExtensions>`, which `dataEntityViewExtensionXml.ts:37` names as part of an `add-field` route. Not defaulted to `AutoReport` — a group the base entity does not have is a compile error. `<FieldGroups>` and `<FieldGroupExtensions>` are not interchangeable and only the latter is touched.
- **Parameter contract.** `required: ['fieldName']` is kept (the mapped-field path genuinely has no `fieldType`), but the type-specific half is now enforced in the operation: a half-filled `dataField`-without-`dataSource` binding is refused instead of written, `fieldType` missing on a table says so, and `dataField`/`dataSource` on a table are rejected rather than ignored. Previously these fell through to a null bridge result and a generic "required parameters may be missing".
- **Tests.** `tests/tools/addFieldDataEntityExtension.test.ts`, 10 cases, fixture copied from the shipped `CurrencyEntity.Extension.xml`. Confirmed RED against the original insertion logic (2 failures) and green after.

### Budget

The raise did not survive the rebase — measured on `main + this PR`, `d365fo_file` came to **10,149 chars against the 10,100 the PR set**, because #804 had spent part of the same headroom. It is not raised now: `params` is `additionalProperties: true`, so the new parameters cost nothing on the wire — only the prose naming them does, and that was paid for by tightening the `params` description. `d365fo_file` measures 9,89x against the unchanged 9,900 ceiling.

The `params`-vs-flat guidance was a real find and is kept — a strict client does drop undeclared top-level keys, and the resulting "required parameters missing" names the wrong cause. The long version now lives in `renderOpSpec` (error text, not on the ListTools budget); the schema keeps a short "NEST them here".

### Verification

- Full `--no-incremental` bridge build against metamodel 7.0.7858.27 — 0 warnings / 0 errors, attestation refreshed. (No CI job compiles the bridge, so this is the only check these branches get.)
- `tsc --noEmit` clean.
- Full vitest: 2488 passed. Two files time out under full-suite load and pass in isolation; a baseline run of `main` on this VM times out on a *different* file (`apiSymbols`), so it is machine load, not this change.

Not done on the VM: an end-to-end `add-field` against a real `AxDataEntityViewExtension` through the running bridge — worth doing before merge if you have the sandbox model handy, since the bridge branch is the one path the repo tests cannot cover.

dynamics365ninja:
## End-to-end run on the VM — found and fixed a blocker in my own rework

I said in the previous comment that the one thing still missing was an end-to-end `add-field` against a real `AxDataEntityViewExtension` through the bridge. I ran it. It failed immediately, and the bug was mine, not the original PR's.

**`a04a583` — the single-op `addField` RPC dropped the three new parameters.**

`RequestDispatcher` dispatches `addField` from two places:

1. the single-op RPC `case "addfield":` in `Dispatch()`, and
2. the `case "addfield": / case "add-field":` arm inside `HandleBatchModify()`.

The rework updated only (2). But `BridgeClient.addField()` — and therefore every `d365fo_file` modify request — goes through (1), which kept calling the 6-argument `AddField` overload. `dataField` / `dataSource` / `fieldGroupName` were dropped before `AddField` ever saw them, so with no binding present it fell through to the table branch and every real request died with:

```
-32602: Table or table-extension 'CurrencyEntity.ConE2EExtension' not found
```

Worth being blunt about why the repo tests were green through all of this: **they mock `BridgeClient`, so they assert the arguments Node *sends*, never the ones C# *reads*.** No amount of TypeScript-side testing could have caught it. That is exactly the gap an end-to-end run exists to close.

`tests/bridge/addFieldDispatchParity.test.ts` is now the standing guard — it asserts both dispatch sites forward the same parameter set, and fails if a third site appears. Verified red before the fix, green after.

### What the end-to-end run then proved

Bridge built `--no-incremental` against metamodel `7.0.7858.27`, driven directly over its stdio protocol, against a scratch `CurrencyEntity.ConE2EExtension` in the Contoso sandbox model (removed afterwards):

| # | Case | Result |
|---|------|--------|
| 1 | mapped field + new field-group extension | ✅ `IMetaDataEntityViewExtensionProvider.Update` |
| 2 | same call again | ✅ `skipped: true` |
| 3 | `dataField` without `dataSource` | ✅ refused |
| 4 | `dataSource` without `dataField` | ✅ refused |
| 5 | second mapped field into the **existing** group | ✅ appended, no duplicate group |
| 6 | unknown extension | ✅ refused with the data-entity message, not the table one |

The XML the metadata API wrote:

```xml
<FieldGroupExtensions>
	<AxTableFieldGroupExtension>
		<Name>AutoReport</Name>
		<Fields>
			<AxTableFieldGroupField><DataField>ConIsoCode</DataField></AxTableFieldGroupField>
			<AxTableFieldGroupField><DataField>ConIsEuro</DataField></AxTableFieldGroupField>
		</Fields>
	</AxTableFieldGroupExtension>
</FieldGroupExtensions>
...
<Fields>
	<AxDataEntityViewField xmlns="" i:type="AxDataEntityViewMappedField">
		<Name>ConIsoCode</Name>
		<Label>@SYS12345</Label>
		<DataField>CurrencyCodeISO</DataField>
		<DataSource>Currency</DataSource>
	</AxDataEntityViewField>
	<AxDataEntityViewField xmlns="" i:type="AxDataEntityViewMappedField">
		<Name>ConIsEuro</Name>
		<Mandatory>Yes</Mandatory>
		<DataField>IsEuro</DataField>
		<DataSource>Currency</DataSource>
	</AxDataEntityViewField>
</Fields>
```

That matches the shipped `ApplicationSuite/Foundation/AxDataEntityViewExtension/CurrencyEntity.Extension.xml` shape element for element — and it matches what the TS direct-XML fallback in `buildAxDataEntityViewFieldXml` emits, so the two writers cannot drift.

**Cold round-trip check.** Element order is load-bearing here: the metadata deserializer silently drops children it meets out of order, so a file that *looks* right can still be lossy. A brand-new bridge process (empty provider cache) re-read the file from disk and found both fields — the order the API wrote survives the round trip.

### Test status

- `tests/bridge/` + `addFieldDataEntityExtension` + `toolSchemaBudget` — **67 passed**, 11 files
- `tsc --noEmit` clean
- `npm run bridge:build` clean, attestation refreshed (metamodel `7.0.7858.27`, 9 sources)

The VM fixture was rolled back; `Contoso/AxDataEntityViewExtension/` is back to holding only `Delta`.

PR #811, opened 2026-08-04
docs: correct architecture drift — the bridge is not the sole write path

dynamics365ninja:
## What

Refresh of the solution architecture diagram (`docs/img/solution-architecture-diagram.svg`) and the docs around it. Last refreshed 2026-06-29; the **structure was still accurate** — no layers missing, nothing gone, layout clean — but the numbers had drifted and one load-bearing claim was wrong.

## Why it matters

The diagram and `ARCHITECTURE.md` both advertised the C# bridge as the **"sole write path"**, with *"no XML string manipulation ever touches metadata files"*. That is not true today:

- Seven `directXml*` writers in `modifyD365File.ts` (`ReplaceCode`, `ModifyProperty`, `AddMenuItemToMenu`, `AddControl`, `AddIndex`, `AddDataEntityExtensionField`, `DeleteAction`), plus `dataEntityViewExtensionXml.ts`.
- Only **13 of the 39** create `objectType`s route to the bridge (`BRIDGE_CREATE_TYPES`).
- `bridgeAdapter.ts:1196` already documents this: *"No C# op backs these — they are served by a direct-XML writer."*

This is the claim the write path's credibility rests on, so it needed more than deleting the word "sole". `ARCHITECTURE.md` now explains **why** the second path exists: the bridge's generic `properties: Dictionary<string,string>` channel cannot carry the structured collections `security-privilege`/`duty`/`role` and `query`/`view` need (EntryPoints, Privileges, Duties, query data sources), so a bridge create would "succeed" and emit a functionally broken object. The XML generators build them correctly instead.

The distinction is about **correctness, not safety** — both paths pass the same grounding gates and the same path-containment check, and the XML writers are structured builders with ambiguity guards (they refuse to guess when a target tag matches more than once), not blind string replacement. That's now stated explicitly so the nuance doesn't read as a hole.

## Counts corrected

Each verified against the source, not against the previous docs:

| Claim | Was | Now | Source |
|---|---|---|---|
| Bridge modify ops | 25 | **31** | `BRIDGE_MODIFY_OPS`, `bridgeAdapter.ts:1189` |
| BP syntax rules | "13 + mined" | **11 static + 4 data-driven** | `validateXpp.ts` (XML002–XML005 are the mined ones) |
| Form pattern rules | FP001–FP010 | **FP000–FP010** | `formPatternValidator.ts:536` |
| Tests | 1400+ | **2500+** | `npx vitest list` collects 2516 |
| TypeScript | 6 | **7** | `package.json` has `^7.0.2` |
| Bridge create types | 13 | **13 (unchanged)** | `BRIDGE_CREATE_TYPES` — already correct |

The create-types number is worth flagging for the reviewer: the C# dispatcher accepts 18 `objectType` strings, so it *looks* stale, but the authoritative gate is the TS routing set and that is 13. The extra 5 are reachable in C# but deliberately never routed there.

## Eval loop

Was missing from the diagram entirely. `eval-gate.yml` now runs bridge attestation, golden regression, knowledge audit and the coverage matrix on every PR, and the README carries two coverage badges with no counterpart in the picture. Added as a third line in the **Automation** box and to the CI/CD row of the tech-stack table.

I did **not** give it its own panel — that would mean reflowing the right-hand column, and the diagram describes runtime + build-time architecture whereas the eval gate is purely CI. Happy to promote it to a full block if reviewers prefer.

## Also

- `ARCHITECTURE.md` referenced `generate_smart_form`; that's the internal name, the exposed tool is `generate_object`.
- README "Safe metadata writes" capability row carried the same overstated no-string-replacement claim.
- Added an `XML writers` node to the bridge mermaid graph.

## Verification

- SVG geometry checked in-browser via `getBBox()` on every edited text node — all still fit their containers. Tightest is `eval gate` (right edge 1248.5 vs box 1276; bottom 378.1 vs box 388).
- The **Automation** box grew 74→82 px; the two dashed database arrows moved from `y=380` to `y=388` so they no longer start inside it. Still 16 px clear of the panel edge.
- Diagram renders correctly; tag balance verified.
- `tests/packaging` + `tests/cli/copilotFiles.test.ts` pass (18 tests) — the only suites that touch docs/README. Full suite not run; changes are documentation-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@
 [![npm](https://img.shields.io/npm/v/d365fo-mcp.svg?logo=npm&color=cb3837)](https://www.npmjs.com/package/d365fo-mcp)
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 [![Node.js](https://img.shields.io/badge/node-%3E%3D24.0.0-brightgreen.svg)](https://nodejs.org/)
-[![TypeScript](https://img.shields.io/badge/TypeScript-6.0-blue.svg)](https://www.typescriptlang.org/)
-[![Tests](https://img.shields.io/badge/tests-1400%2B-brightgreen.svg)](docs/TESTING.md)
+[![TypeScript](https://img.shields.io/badge/TypeScript-7.0-blue.svg)](https://www.typescriptlang.org/)
+[![Tests](https://img.shields.io/badge/tests-2500%2B-brightgreen.svg)](docs/TESTING.md)
 <!-- coverage-badge:start -->
 [![Core coverage](https://img.shields.io/badge/core_coverage-100%25-brightgreen.svg)](eval/COVERAGE.md) [![Total coverage](https://img.shields.io/badge/total_coverage-100%25-lightgrey.svg)](eval/COVERAGE.md)
 <!-- coverage-badge:end -->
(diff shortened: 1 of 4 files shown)


dynamics365ninja:
Follow-up pushed: the eval loop now has its own **EVAL LOOP** panel instead of the single line in the Automation box.

It takes the empty right-column slot under INDEX PIPELINE (`1044..1296, 424..662`), so the canvas keeps its `1320x940` box — no reflow of the left column. The panel shows the real topology from `AGENT_EVAL_LOOP.md` §3: implementer (VM, grounded tools only) → corpus (NDJSON, one record per run) → improver (repo, reproduce → fix → PR), with a violet dashed arrow closing back into the MCP server core.

Three things had to move to free the slot, which is the part worth a reviewer's eye:

- the two dashed *produces both databases* arrows ran at `x=1240/1248`, straight through where the panel now sits → rerouted through the right margin at `x=1304/1312`, turning below the pipeline panel at `y=412/418`
- `produces both databases` → `(560, 668)` on the long horizontal run
- `reads AOT XML / at index build` → `(1024, 706/718)`, right-anchored, clear of the DATA SOURCES edge it would otherwise have sat on

Legend gained a third entry (*improvement flow*), box widened `404→640`. Header gained a fourth stat chip tying the panel to the core-coverage badge already in the README.

`ARCHITECTURE.md` gained a matching **Self-improving eval loop** section so the prose doesn't lag the picture — the two oracles (golden metadata primary; SysTest for behaviour a golden cannot see), fixtures vs case outputs, per-run isolation, and what the K/E/T coverage rule actually counts.

Verification: 80 cases in `eval/cases`, coverage core 44/44 and total 78/78 read from `eval/coverage.json`. Geometry checked in-browser — pairwise `getBBox()` across all 123 text nodes reports **zero** collisions, every panel text is inside its own rect, nothing crosses the canvas bounds.

PR #815, opened 2026-08-06
fix(workspaceDetector): stop silently picking the wrong .rnrproj when multiple projects exist

laeliand:
detectD365Project() previously defaulted to projectFiles[0] (alphabetically first) whenever a workspace contained more than one .rnrproj and no unique signal disambiguated them, causing addToProject to silently register new files into an unrelated project. It now only auto-resolves when exactly one .rnrproj exists, or one uniquely matches the workspace folder name; otherwise it returns null so the caller must pass projectPath explicitly.

createD365File.ts's "no projectPath could be resolved" warning (previously duplicated 3x with generic text) now lists every detected candidate when the workspace is ambiguous, and the projectPath schema description spells out that it is effectively mandatory in multi-project workspaces.

Adds regression tests for detectD365Project covering the single-project, ambiguous-multi-project, name-match-resolved, and no-project cases.

### DIFF
--- a/src/tools/createD365File.ts
+++ b/src/tools/createD365File.ts
@@ -73,7 +73,11 @@ async function withProjectFileLock<T>(projectPath: string, fn: () => Promise<T>)
  * wrong project.
  */
 function buildNoProjectPathWarning(): string {
-  const candidates = getConfigManager().getAllDetectedProjects();
+  // The WORKSPACE candidates, not getAllDetectedProjects(): under
+  // D365FO_SOLUTIONS_PATH the latter lists every project across every solution,
+  // which would put a wrong count behind "in this workspace" and can run to
+  // dozens of lines in what should be a short, actionable warning.
+  const candidates = getConfigManager().getWorkspaceProjectCandidates();
   if (candidates.length > 1) {
     return `\n⚠️ addToProject=true but no projectPath could be resolved: ${candidates.length} .rnrproj ` +
       `files were found in this workspace and none matched unambiguously.\n` +
--- a/src/tools/createD365File.ts
+++ b/src/tools/createD365File.ts
@@ -82,7 +82,13 @@ function buildNoProjectPathWarning(): string {
       candidates.map(c => `   - ${c.modelName}: ${c.projectPath ?? '(no .rnrproj)'}`).join('\n') + '\n';
   }
   return `\n⚠️ addToProject=true but no projectPath could be resolved.\n` +
-    `Add projectPath to .mcp.json or pass it as a parameter.`;
+    `The file was created on disk but was NOT added to any Visual Studio project.\n\n` +
+    `Pass projectPath as a parameter, or add it to your .mcp.json:\n` +
+    `  {\n` +
+    `    "servers": { "context": {\n` +
+    `      "projectPath": "K:\\\\VSProjects\\\\YourSolution\\\\YourModel\\\\YourModel.rnrproj"\n` +
+    `    } }\n` +
+    `  }\n`;
 }
 
 const CreateD365FileArgsSchema = z.object({
@@ -133,13 +139,9 @@ const CreateD365FileArgsSchema = z.object({
     .string()
     .optional()
     .describe(
-      'Path to .rnrproj file. Effectively MANDATORY, not optional, in any workspace with more than one ' +
-      'D365FO project/model: auto-detection can only resolve unambiguously when exactly one .rnrproj is ' +
-      'found (or one uniquely matches the workspace name). Whenever a workspace has multiple .rnrproj ' +
-      'files, auto-detection deliberately refuses to guess between them — call get_workspace_info first ' +
-      'to see the detected candidates, then pass this parameter explicitly on every create call so the ' +
-      'file is registered into the intended project instead of being silently added to the wrong one or ' +
-      'not added at all.'
+      'Path to .rnrproj file. Required for addToProject in any workspace holding more than one .rnrproj: ' +
+      'auto-detection resolves a project only when there is exactly one, or one whose folder matches the ' +
+      'workspace name, and otherwise refuses to guess. Call get_workspace_info to list the candidates.'
     ),
   solutionPath: z
     .string()
--- a/src/tools/createD365File.ts
+++ b/src/tools/createD365File.ts
@@ -63,6 +63,28 @@ async function withProjectFileLock<T>(projectPath: string, fn: () => Promise<T>)
   }
 }
 
+/**
+ * Builds the "no projectPath could be resolved" warning shown when
+ * addToProject=true but neither projectPath/solutionPath nor auto-detection
+ * produced a usable path. When multiple .rnrproj files exist in the workspace,
+ * auto-detection deliberately refuses to guess between them (see
+ * workspaceDetector.ts detectD365Project) — list the candidates so the caller
+ * knows to pass projectPath explicitly instead of silently landing on the
+ * wrong project.
+ */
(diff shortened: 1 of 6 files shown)


dynamics365ninja:
## Review

The diagnosis is right and the direction is right — `projectFiles[0]` was never a defensible default. But the refusal is applied one layer too low, and the headline improvement (listing candidates) does not fire in the scenario the PR describes. Requesting changes; details below, roughly in priority order.

Verified on the branch: `tsc --noEmit` clean, full suite green.

### 1. The candidate list is empty exactly where it matters

`buildNoProjectPathWarning()` reads `getConfigManager().getAllDetectedProjects()`. That array is populated **only** by the `D365FO_SOLUTIONS_PATH` scan (`configManager.ts:71`, `:106`, `:224` — its own doc comment says so).

So in the scenario this PR is about — a workspace with 2+ `.rnrproj` and no `D365FO_SOLUTIONS_PATH` — the list is `[]`, `candidates.length > 1` is false, and the user gets the short fallback string. The detailed listing only appears for users who *do* have `D365FO_SOLUTIONS_PATH` set, i.e. the setup where detection usually resolved a primary anyway.

The real candidates exist in `projectFiles` inside `detectD365Project` and are discarded. They need to be surfaced from there rather than read back from an unrelated scan.

### 2. Returning `null` pushes the decision into worse fallbacks

`detectD365Project` is not addToProject-specific — it is the model-name source (`configManager.ts:842`). On `null`, `autoDetectD365Project` continues down the chain to Priority 4, which **still does `files[0]`** (`workspaceDetector.ts:265`) over `K:\VSProjects`, `C:\Projects`, `Documents\Visual Studio 2022\Projects`.

Net effect: instead of arbitrarily picking a project *inside the user's workspace*, the server can now arbitrarily pick one from a **completely different solution root** — a strictly worse silent-wrong-project outcome. And once Priority 4 supplies that bogus project, the `!this.autoDetectedProject` guard at `configManager.ts:240` suppresses the `D365FO_SOLUTIONS_PATH` primary selection, including its Microsoft-demo-model skipping.

If "never guess" is the rule, it has to apply to `workspaceDetector.ts:265` and `configManager.ts:241` in the same change.

### 3. "Ambiguous" is the normal VS 2022 state, not an edge case

`configManager.ts:355-357` notes that VS 2022 *always* sends the solution root, the ancestor of all projects. So for every multi-project solution, `detectD365Project(root)` now returns `null` — losing `modelName`, not just `projectPath`, which degrades far more than `addToProject`.

Same for Priority 3b (`workspaceDetector.ts:234`): the branch documented as "recommended for multi-solution setups" now returns `null` by construction whenever the root holds more than one solution, and logs a `console.error` for a configuration that is working as intended. Ditto the packagePath fallback (`configManager.ts:173`), which scans `PackagesLocalDirectory`-shaped paths that routinely contain many models.

Suggested shape: don't collapse "which model" and "which .rnrproj" into one all-or-nothing `null`. Return `D365ProjectInfo` with `projectPath`/`solutionPath` undefined plus `ambiguousProjects: string[]`. `addToProject` then correctly refuses *and* can list candidates (fixes 1), while model detection keeps working. Bonus: when all candidates share the same `<Model>`, the model is unambiguous even though the project isn't.

### 4. The demo-model rescue narrowed, and returns a mismatched pair

The block at `workspaceDetector.ts:148-162` is now only reachable when `nameMatch` succeeded, since every other multi-project path returns earlier.

Also — pre-existing, but sitting in the block this PR restructures — `modelName` is read at line 150 from the *original* `primaryProject` and never re-read after `primaryProject = pf`, so the rescue returns `{ projectPath: <the good project>, modelName: 'FleetManagement' }`. Worth fixing while here.

### 5. Oversized tool description, and advice that can't be followed

The seven-line `projectPath` description ships in every `tools/list` payload, for the server's busiest tool. It tells the model to "call get_workspace_info first to see the detected candidates" — but that listing (`toolHandler.ts:479`) is again gated on `allDetectedProjects.length > 1`, i.e. `D365FO_SOLUTIONS_PATH` again. Suggest trimming to two sentences and making the claim true.

### 6. Lost remediation detail

The third call site lost its concrete `.mcp.json` example in favour of "Add projectPath to .mcp.json or pass it as a parameter" — and per (1) that short branch is the one that actually fires most of the time.

### 7. stderr noise

`detectD365Project` runs up to 4x per detection cycle; each ambiguous root now emits a multi-line `console.error`. This should be `debugLog`, or emitted once per decision.

### Tests

Good — real temp dirs instead of mocking fs (right call, the behaviour under test *is* tree traversal), cross-platform, clean teardown, and the name-match case is set up correctly. Gaps: nothing covers `buildNoProjectPathWarning` (where the actual defect in (1) lives), nothing pins that `autoDetectD365Project` doesn't fall through to the well-known dirs on ambiguity (2), and the demo-model path is untested.

laeliand:
Thanks Filip — really thorough review, and appreciate you going ahead and pushing the follow-up fix directly rather than just leaving comments. Looks like it covers all the points (ambiguous-model resolution, the well-known-dir fallback, the candidate list actually being populated in the scenario that matters, the demo-model mismatch, the docs/message text).

dynamics365ninja:
Pushed `ee28f33` onto this branch with three follow-ups from review. Direction of the change is right — the silent wrong-project pick was a real bug — but the candidate-listing commit had a way of coming back.

**1. `workspaceProjectCandidates` is now its own field.** Recording the workspace candidates into `allDetectedProjects` gave one field two meanings, and with `D365FO_SOLUTIONS_PATH` set that had teeth:

```js
this.allDetectedProjects = candidates;        // workspace candidates
...
const needsScan = this.allDetectedProjects.length === 0;   // ← now false
if (needsScan) { /* solutions-path scan — skipped */ }
const all = this.allDetectedProjects;                      // ← workspace candidates
if (all.length > 0 && !this.autoDetectedProject && !isNowStale) {
  const primary = all.find(p => !isMicrosoftDemoModel(p.modelName)) ?? all[0];
  this.autoDetectedProject = primary;         // ← first workspace .rnrproj, silently
}
```

So the solutions-path scan is skipped as already done, and its "no project detected → use the first one found" fallback adopts an arbitrary workspace `.rnrproj` — the exact bug this branch removes, reinstated one branch over. It also races `initEagerScan()`, which assigns that field unconditionally, so which list survives depends on timing; the comment "never overwrites a solutions-path scan" holds in one direction only.

**2. `detectD365Project` now resolves the project when exactly one candidate carries the single custom model.** The branch refused as soon as folder-name matching failed — but by that point the models have been read and they say which file is the real one, the others being Microsoft demo projects the VS wizard named and nobody renamed. That is an ordinary workspace (a leftover tutorial project next to the real one), the old code got it right via the demo-model rescue, and refusing silently stopped `addToProject` working there. Several projects sharing one model still yield no `projectPath`, as intended.

**3. The well-known-roots scan now applies that rule at least as strictly**, as its comment already claimed. It was still doing `files.find(f => rootModels.get(f) === customModels[0])` — picking the first of several projects carrying one model, in the very situation `detectD365Project` refuses. Out there the root is not the user's workspace, so a wrong pick lands in an unrelated solution; it should be the stricter of the two, not the looser.

Also: `buildNoProjectPathWarning()` read `getAllDetectedProjects()`, which under `D365FO_SOLUTIONS_PATH` spans every solution — a wrong count behind "in this workspace", and potentially dozens of lines in a warning meant to be short. It now reads the workspace candidates.

Two new tests: the two lists staying separate under `D365FO_SOLUTIONS_PATH`, and the demo-plus-real workspace resolving. Full suite green (195 files, 2583 passed).

PR #841, opened 2026-08-07
feat(get_object_info): plural objects[] form, retire batch_get_info

### LINKED ISSUE #831: get_object_info has no multi-object form and batch_get_info is never chosen — 13 sequential lookups in one session
Found in the session audit (#824).

`get_object_info` was called **13 times** in one session, one object per call. `batch_get_info` — which exists precisely for this, and was advertised in the same tool list — was called **zero times**.

Server-side cost of those 13 calls: 10,0 s. Agent-side cost: 13 round trips ≈ **39 AIU**. The tool is fast; the round trips are what the session pays for.

Several of them were provably batchable — consecutive calls in adjacent turns against related objects:

```
19:11:51  Reading table <Table>          ┐ same object, two types,
19:11:51  Reading form  <Table>          ┘ issued together — this pair the agent did batch
19:12:10  Reading table <Table>          ┐
19:12:10  Reading form  <Table>          │ post-bridge-ready retry of the same three
19:12:10  Reading table-extension <Table>.<Model>Extension  ┘
19:13:28  Reading class <Class>FormHandler
19:13:35  Reading class <OtherModel><Class>FormHandler
19:16:31  Reading class <Table1><Model>_Extension           ┐
19:16:35  Reading class <Table2><Model>_Extension           │ three sibling CoC classes,
19:16:35  Reading class <Table3><Model>_Extension           ┘ three separate turns
```

## Why the agent never reaches for `batch_get_info`

Two plausible causes, both fixable:

1. **Discovery.** With 105 tools the host was in search-based tool mode (see #825). `batch_get_info` appears in one of the agent's `tool_search` queries and still never got used — it was found and passed over.
2. **Framing.** `get_object_info` (1 720 B, 850 B of description) reads as *the* way to read an object. `batch_get_info` (1 047 B, 229 B of description) reads as a specialised variant. Given a name to look up, the obvious tool wins every time.

## Suggested fix

Collapse the two. Give `get_object_info` a plural form and retire `batch_get_info`:

```json
{"objects":[{"objectType":"class","objectName":"<Table1><Model>_Extension"},
            {"objectType":"class","objectName":"<Table2><Model>_Extension"},
            {"objectType":"class","objectName":"<Table3><Model>_Extension"}]}
```

Keep `{objectType, objectName}` working as a one-element shorthand. This removes a tool from the surface (−1 047 B per request), removes the choice the agent keeps getting wrong, and makes the batched path the *default* path rather than an alternative that must be discovered.

If the two must stay separate, then `get_object_info`'s description needs a first line the agent cannot miss — *"For more than one object use `batch_get_info` — it is one round trip instead of N"* — but merging is the more reliable fix. The same reasoning applies to `search` vs `batch_search` if that pair is exposed.

Expected saving on the audited session: 13 calls → ~5, roughly **−24 AIU and −30 s**.

## Acceptance

- `get_object_info` accepts `objects[]` and returns per-object sections in one result
- single-object form unchanged
- an eval case with 3+ objects asserts one tool call, not N

dynamics365ninja:
Collapses `batch_get_info` into `get_object_info` so the batched read is the **default** path, not a discoverable alternative the model keeps passing over.

## What changed

| Area | Change |
|---|---|
| `src/tools/getObjectInfo.ts` | New plural form `objects: [{objectType, name, options?}]` (min 1, max 10). Both forms normalise to the same `ObjectRef[]`; 1 ref → plain single-object result, 2+ → parallel fan-out with per-object sections. |
| `src/tools/batchGetInfo.ts`, `src/server/toolSchemas/batchGetInfo.ts` | Deleted. Their fan-out (`Promise.all` over `READER_DISPATCH`), `## N. Name [TYPE]` section format, `Fetched: N object(s) in parallel \| Success: x/N \| Time: Nms` header and `withNotFoundGuidance` wiring were **moved** into `getObjectInfo.ts`, not re-written. |
| `src/tools/objectInfoRegistry.ts` | `BATCH_INFO_TYPES` removed (no consumer left), along with the three `security-*` `READER_DISPATCH` entries it existed for. |
| `src/server/toolSchemas/getObjectInfo.ts` | `objects[]` added; top-level `required` dropped (both forms are valid — the same shape `search` uses for `query` vs `queries[]`); description leads with the plural form. |
| Surface bookkeeping | `toolHandler.ts` (import + `case`), `toolAnnotations.ts`, `transport.ts` `FAST_TOOLS`, `toolProgressMessage.ts` (plural branch), `src/index.ts` catalog + category string (`2 discovery` → `1 discovery`), `systemInstructions.ts`, `.github/copilot-instructions.md`, `.claude/agents/eval-implementer.md`. |
| Docs | Tool count 26 → 25 in `README.md`, `docs/MCP_TOOLS.md` (header, server-modes line, section count), `docs/ARCHITECTURE.md`, `docs/QUICK_START.md`, `docs/MCP_CONFIG.md`, `docs/SETUP.md`. |

## Acceptance criteria

**1. `get_object_info` accepts `objects[]` and returns per-object sections in one result** — met.
`{objects:[{objectType:"class",name:"A"},…]}` runs every lookup through `READER_DISPATCH` in parallel and returns a single `content[0]` with a `Fetched: 3 object(s) in parallel | Success: 3/3` header and `## 1. A [CLASS]` … sections. Per-object failures are reported inline without failing the call; `isError` is true only when every lookup failed. Max 10, the same ceiling `batch_get_info` had.

**2. Single-object form unchanged** — met (this is the compatibility guarantee).
`{objectType, name}` still returns the plain reader result — no batch header, no numbered sections — and the top-level `compact` / `methodOffset` shortcuts, `options` passthrough, the `options.members:"names"` code-completion fold and the not-found guidance all behave as before. A one-element `objects[]` is asserted identical to the single form. A top-level `options` on a plural call applies to every entry; a per-entry `options` wins.

**3. Eval case with 3+ objects asserting one tool call, not N** — met, with a caveat.
`eval/cases/L2-batched-object-reads.json`: three sibling table extensions (VendGroup, PaymTerm, InventItemGroup) whose instruction mandates resolving all three base tables in a **single** `get_object_info(objects=[…])` call and states that the run record must show exactly ONE call, not three. **`golden_pending: true`** — the golden is a real build artifact and can only be captured on the VM (repo convention, §6.4; same shape as `L4-bridge-drops-*`). Because the case schema has no field for asserting tool-call counts, the assertion lives in the instruction and is pinned by `tests/eval/batchedObjectReadsCase.test.ts` so it cannot be edited away silently; the behavioural half (3 objects → 1 call → 3 sections, 3 reader invocations) is asserted VM-free in `tests/tools/getObjectInfoPlural.test.ts`.

## Tool count

**26 → 25.** Updated in `src/index.ts` (catalog + `1 discovery + …` category string), `tests/utils/toolInventory.test.ts` (`toHaveLength(25)` for both inventories, non-local `16` → `15`, and `batch_get_info` added to the retired-names guard) and `tests/utils/toolSchemaBudget.test.ts` (`tools.length` 25).

## `TOTAL_BUDGET`

**Not touched** (still `63_300`). The ListTools payload went 63,258 → 63,213 chars — a net saving of only 45, because `get_object_info` grew 1,720 → 2,722 B while `batch_get_info`'s 1,047 B went away. I deliberately kept the full `objectType` enum repeated inside `objects[].items` (~330 B) to match how `search` repeats its type enum inside `queries[]` — nested enums are what make the model fill the field correctly — rather than squeezing the shared budget unilaterally. Headroom is now 87 chars, up from 42. The stale "Current actual ≈ 63,175 / d365fo_file ≈ 9,6xx" prose in that test's comment was left alone to avoid conflicting with sibling PRs.

## Tests

- **New** `tests/tools/getObjectInfoPlural.test.ts` (13 tests): 3-objects-in-one-call, mixed types with the right per-reader argument key, partial failure, not-found guidance per object, option precedence, empty/`>10` rejection, "neither form" rejection, single-form result shape, one-element equivalence, class shortcuts, published-surface check (`batch_get_info` gone, `objects[]` advertised, no global `required`), progress-message labelling.
- **New** `tests/eval/batchedObjectReadsCase.test.ts` (4 tests): case spec well-formedness, 3+ objects of readable types, the one-call mandate, no reference to the retired tool.
- **Removed** `tests/tools/batchGetInfo.test.ts` — every one of its assertions is ported to the plural tests.
- **Updated** `tests/tools/object-extension-info.test.ts` (dropped the `BATCH_INFO_TYPES` assertion).
- Full suite: **206 files, 2717 passed, 9 skipped**. `npx tsc --noEmit` clean.

## Deliberate omissions / follow-ups

- **`search` vs `batch_search`** — the issue muses about applying the same reasoning to that pair. Out of scope here and untouched; `search` has already absorbed its batch form as `queries[]`, so the equivalent collapse is effectively done. Flagged for the record.
- **Security artifact types** — `batch_get_info` accepted `security-privilege` / `security-duty` / `security-role`; `get_object_info` never did. Rather than widen the enum (and duplicate an existing tool), those three left with it: `security_info(mode="artifact")` is and remains their reader. `BATCH_INFO_TYPES` and the three `READER_DISPATCH` entries are removed as now-dead code.
- **Live eval run** — `L2-batched-object-reads` needs an implementer run on the VM to capture its golden and flip `golden_pending` to false.

Closes #831


dynamics365ninja:
Pushed `d636fe6` — three follow-ups.

### 1. `objects[]` entry key `name` → `objectName`

Unified with the other plural object lists on the server: `verify_d365fo_project` already takes `objects:[{objectType, objectName}]` and `run_bp_check` (#839) adopts the same key. Renamed in the tool schema (`items.properties` + `items.required` + the description), the zod `ObjectRefSchema` and the `toObjectRefs` normalisation, the progress-message plural branch, `systemInstructions.ts`, `.github/copilot-instructions.md`, `docs/MCP_TOOLS.md`, `eval/cases/L2-batched-object-reads.json` and both test files.

**No `name` alias.** The form is unreleased, so there is nothing to keep working, and an alias is permanent ListTools weight for a spelling nothing has ever sent. A misspelled entry fails argument validation with `objectName` named in the message — pinned by a new test, plus an assertion that `items.properties.name` is undefined.

⚠️ **One correction to the framing.** The top-level single-object parameter is `name`, **not** `objectName` — the plural entry and the singular parameter therefore do *not* end up sharing a spelling. Renaming the top-level parameter would break a shipped surface, so it is out of scope here; flagging it as a separate decision. `search`/`batch_search` is the other precedent and it goes the other way: `queries[]` entries use `query`, mirroring the singular `query`. Left untouched — out of scope.

### 2. Eval gate

`eval/coverage.json` + `eval/COVERAGE.md` regenerated (`L2-batched-object-reads` was added without rerunning the matrix, so `eval:coverage --check` was failing). The case lands in `orphans.cases` — no coverage leaf claims it — alongside the three cases already there.

Full gate, in workflow order:

| Step | Result |
|---|---|
| `npm run build` | ✅ |
| `npx vitest run` | ✅ 206 files, **2725 passed**, 2 skipped |
| `npm run eval:knowledge-audit` | ✅ 285 references covered |
| `npm run eval:coverage -- --check` | ✅ core 44/44, total 78/78 |

`npx tsc --noEmit` clean.

### 3. `TOTAL_BUDGET` re-measured

| | before | after |
|---|---|---|
| ListTools payload | 63,213 | **63,255** (+42) |
| largest tool (`d365fo_file`) | 9,888 | 9,888 (unmoved) |

`TOTAL_BUDGET` stays **63,300** and `LARGEST_TOOL_BUDGET` stays **9,900** — both are still the next round hundred above the measurement, which is how the constants are already used ("headroom is small on purpose so creep is caught early"): 45 chars over the total, 12 over the largest tool. `get_object_info` is ~2.8 KB, nowhere near the per-tool cap, so the rename could not move it. What *was* wrong is the prose: the comment claimed `≈ 63,175` / `d365fo_file ≈ 9,6xx`, both stale. Replaced with the real figures and a line recording this rename's +42.

PR #899, opened 2026-08-12
fix(doctor): detect stale legacy .env overrides for UDE path settings

laeliand:
customPackagesPath/microsoftPackagesPath/packagePath are meant to be derived live from the active XPP config on UDE, so a platform update is picked up automatically. But the setup wizard still writes these into an instance's legacy .env at setup time, and once  deleted the old framework/model-store folder, that hardcoded .env value keeps outranking the (now-correct) XPP-config auto-detection at every startup — see configManager.ts's envContext, which checks these env vars before ever consulting the XPP config. The bridge then dies silently with "C# bridge unavailable (ude)", with nothing pointing at the actual cause.

doctor's existing legacyEnvChecks only flags a *disagreement* between .env and the JSON config — it stays silent when the JSON config has no value at all for a setting, which is the normal (and recommended) state for these three on UDE, so the stale-but-unopposed .env value was invisible to it.

Add settingSource() to tell "explicitly configured" apart from "inherited from .env", and extend doctor's path-existence checks (previously packagePath only) to customPackagesPath and microsoftPackagesPath too, with a fix message that points at the actual cause (a stale, legacy .env pin) instead of suggesting to re-pin the same kind of static value.

Verified against two real local instances: correctly reports OK for one whose .env was already fixed, and catches the same bug live on a second instance still pinned to a deleted platform-version folder.

### DIFF
--- a/src/cli/commands/doctor.ts
+++ b/src/cli/commands/doctor.ts
@@ -13,7 +13,7 @@ import { bridgeBuildCommand, dataRoot, installMode, isWindows, paths, repoRoot }
 import { commandExists } from '../exec.js';
 import { isLegacyInstanceLayout, listInstances, type Instance } from '../instances.js';
 import { checkRelease } from '../npmRegistry.js';
-import { conflictingLegacyValues, readPath, readSetting, type SettingsStore } from '../settingsStore.js';
+import { conflictingLegacyValues, readPath, readSetting, settingSource, type SettingsStore } from '../settingsStore.js';
 import { instanceTarget, rootTarget, type Target } from '../target.js';
 import { isXppConfigStale, listXppConfigs, xppConfigDir } from '../xppConfig.js';
 import { describePackagesRootScan, packagesRoots } from '../../utils/packagesRoot.js';
@@ -98,7 +98,8 @@ function checkConfig(target: Target, label: string): CheckResult {
 function checkPackagesRoot(store: SettingsStore, label: string): CheckResult[] {
   if (!isWindows) return [];
 
-  const configured = String(readSetting(store, settingByPath('environment.packagePath')!) ?? '').trim();
+  const setting = settingByPath('environment.packagePath')!;
+  const configured = String(readSetting(store, setting) ?? '').trim();
   const detected = packagesRoots();
 
   if (!configured) {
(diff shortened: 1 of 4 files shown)


dynamics365ninja (commented):
Good catch on the underlying failure — a stale `.env` silently outranking live XPP-config detection is exactly the kind of thing that produces "C# bridge unavailable (ude)" with nothing to go on, and `settingSource()` is the right primitive for telling it apart from a config typo. The `legacyEnvChecks` gap you identified (silent when the JSON config has no opposing value, which is the normal state for these three) is real.

I audited it against the codebase and found three problems, all from the same cause: `missingPathFix` is shared by three settings that are not the same kind of thing. I've pushed a fixup to this branch (`8d9ebb9`) rather than leaving it as a review round-trip — full detail in the commit message, summary here.

**1. Routing `checkPackagesRoot` through the shared message regressed the existing check.**

A traditional VM with a stale `D365FO_PACKAGE_PATH` in `.env` — the normal state after a wizard-written `.env`, per your own PR description — now gets:

> packages root does not exist (…) — pinned by legacy .env (D365FO_PACKAGE_PATH), not the JSON config
> fix: remove D365FO_PACKAGE_PATH from .env; **this UDE value is normally auto-detected from the active XPP config**

On traditional, `packagePath` is not resolved from an XPP config — `packagesRoots()` scans the drives for it (`environment.packagePath`'s own description says so: "Machine-wide on a traditional VM; UDE resolves it from the XPP config instead"). The env branch also drops the two things that made the old message actionable: `found instead: <detected roots>` and `SETUP_COMMAND`.

**2. The fix line proposes a value that cannot be right for the two new settings.**

`checkUdeOverridePath` passes `packagesRoots()` as `detected`, and `missingPathFix`'s fallback text is hardcoded to PackagesLocalDirectory. So a JSON-pinned, missing `microsoftPackagesPath` is told:

> fix: set environment.microsoftPackagesPath to K:\AosService\PackagesLocalDirectory

`<drive>:\AosService\PackagesLocalDirectory` is a traditional-VM artifact; the UDE `FrameworkDirectory` and `ModelStoreFolder` are never that. And on a UDE box with no AosService directory, the no-candidates branch prints "No `<drive>:\AosService\PackagesLocalDirectory` found on any drive (C: to Z: were scanned)" as the explanation for a missing ModelStoreFolder.

**3. "Remove the stale pin" is wrong advice for `customPackagesPath` on a traditional VM.**

This is the one I'd most want a second opinion on, because it inverts a documented workflow. `docs/MCP_CONFIG.md:165-187` explicitly instructs users to set `D365FO_CUSTOM_PACKAGES_PATH` on **traditional** installs when metadata is junctioned into PLD — it is the documented fix for "modify fails with outside package roots". The new check treats any set-but-missing value as a stale UDE override and tells the user to delete it, i.e. to break `modify` on exactly the layout the docs told them to configure that way.

**What the fixup does**

Each path setting now carries facts about itself — how it resolves when nothing pins it (`autoSource(kind)`), what values doctor can actually propose, and what it ought to point at. `autoSource` is the load-bearing one: a stale pin is only *stale* where something else would resolve the value live; where nothing would, the pin **is** the configuration and gets corrected instead of deleted. `environmentKind()` reads `environment.type` and falls back to the same XPP-config detection the server itself uses, so doctor's diagnosis matches the runtime instead of second-guessing it.

Your `.env` diagnosis is preserved everywhere it applies, and `packagePath`'s config-sourced messages are back to their pre-PR wording.

One smaller thing, also in the fixup: `settingSource`'s docblock says it "Mirrors readSetting's own precedence", but it skipped `stripInlineComment`, so `KEY=  # note` reported `'env'` while `readSetting` yields an empty value. Unreachable through doctor today (the empty value returns early), but it would have drifted for the next caller.

**One limitation worth noting in the PR description rather than fixing here.** The runtime override is `process.env` (`configManager.ts:782-783`), and the documented way to set these is the `env` block in `.mcp.json` (`docs/MCP_CONFIG.md:165`), not only a `.env` file. `settingSource` reads the root/instance `.env` only, so a pin arriving that way stays invisible to doctor. Your PR fixes the wizard-written `.env` case, which is the one you hit — just worth saying so explicitly, so the next person doesn't read a clean doctor run as proof that no stale pin exists.

Tests: `missingPathFix` is now pure, so all twelve setting × environment × source combinations are asserted with no Windows box, XPP config or real `.env` needed — including two negatives that pin the regressions above. `tsc --noEmit` clean, 294 files / 3633 tests green, `biome lint` clean, and `npm run doctor` verified on a real traditional VM.

PR #907, opened 2026-08-12
feat(knowledge): add bp-moniker — validate/search/suppress BP-check monikers

laeliand:
A BP-check moniker has been proposed wrong before (issue: an assistant typed one from memory that did not exist, caught only by reading the xppc log by hand), and the existing bp-check skill's fix recipes cover only a handful of monikers by name — anything else was a dead stop asking the user to look it up manually.

Adds a real ground-truth catalog instead of a hand-typed list:
  - scripts/extract-bp-catalog.ps1 scans a local D365FO install for two real sources: every model's AxRuleSet/BPRules.xml (union = 409 canonical moniker names) and the resx-backed resource classes inside the .NET-authored rule DLLs under bin/BPExtensions/*.dll (545 real message/description strings). Output is a generated TS module (src/knowledge/bpMonikers/catalog.generated.ts, 577 entries), following this project's existing embedded-knowledge convention rather than a runtime-loaded JSON file.
  - src/knowledge/bpMonikers/index.ts: validateMoniker (exact, case-insensitive, never fuzzy-corrects a typo into a false positive), searchMonikers (keyword/token overlap against real message+description text, for when there is a scenario but no BP-check finding yet), and buildSuppressionXml (renders a real <Diagnostic> block matching the actual Microsoft AxIgnoreDiagnosticList shape, not just the shorter template comment those files also carry).
  - New get_knowledge(kind="bp-moniker") action, folded into the existing unified tool rather than published standalone (tests/utils/ toolSchemaBudget.test.ts had ~0 chars of headroom; ceiling raised by 1,000 with a matching justification comment, same convention as prior raises).
  - run_bp_check now cross-references every plain-text finding's moniker against the catalog and appends its real description inline (parseBpFindings / renderFindingsSection in runBpCheck.ts) — additive only, existing return shapes are unchanged and all 63 existing tests still pass.

Tests cover all three use cases end-to-end against the real extracted catalog (not synthetic fixtures): catalog sanity, exact-moniker validation including the deliberate-miss case, free-text search including the zero-match case, suppression XML generation including the unknown-moniker warning path, and run_bp_check's finding parsing/cross-referencing.

Known limitation, stated in the module docblock rather than hidden: coverage is uneven. Core X++-authored rules (the bulk of the canonical set) are not known to carry the same resource-class message/description text as the .NET-authored ones — those entries have a confirmed real name but null message/description, which the schema states explicitly rather than leaving ambiguous.


dynamics365ninja:
## Audit + a fix commit pushed to this branch

Heads-up first: I pushed **879f08f** straight onto `feat/bp-moniker-catalog` (fast-forward, no force) rather than opening a separate PR, so the fixes and the review sit together. Shout if you'd rather have had it as a diff to look at — happy to revert and re-send that way.

The approach here is right, and the "extract ground truth instead of typing a list" instinct is the only thing that actually kills this class of bug. What the audit found is all execution, not design — nothing below needed the concept rewritten.

I verified against a local D365FO install (a different one from the 10.0.2527.174 box the catalog came from): every `<Diagnostic>` in all **299** `AxIgnoreDiagnosticList` files was parsed — 19,619 blocks carrying both `<Path>` and `<ElementType>`, **10,915** of them `DiagnosticType=BestPractices`.

### 1. `<Path>` was wrong for almost every element type (this was the blocker)

The real rule is: **path segment = `ElementType` minus its `Ax` prefix, singular.** That holds on **1,447 of 1,447** real BP entries carrying both fields — zero exceptions. The pluralised forms (`Classes`, `Tables`, `ExtendedDataTypes`, `MenuItemsDisplay`) appear in **no** real entry.

That matters more than a cosmetic mismatch: a suppression with a path that matches no element **silently suppresses nothing**. The build keeps emitting the warning and there's no signal saying why — the exact failure mode this PR was written to prevent.

I replayed 418 real top-level BP suppression entries (22 element types, all models) through the generator:

| | correct | wrong path | type not supported |
|---|---:|---:|---:|
| before | 149 | 213 | 56 |
| after | **418** | 0 | 0 |

The 149 that passed were `AxSecurityPrivilege` / `AxSecurityDuty` / `AxSecurityRole` — the three types that happen to have no plural form, and `AxSecurityPrivilege` was the only type the test suite asserted a `<Path>` for. That's why this got through green tests.

`PATH_SEGMENT` is gone; it's `elementType.replace(/^Ax/, '')`.

### 2. The block shape was inverted relative to real files

| element | real occurrence | this PR emitted | now |
|---|---:|---|---|
| `<Justification>` | **95%** (10,384) | never | always |
| `<Message>` | 43% (4,737) | always | only when actually known |
| `<ElementType>` | 13% (1,447) | always | only when supplied |
| `<ItemSpecific>` | 9% (999) | always | opt-in |

Microsoft's own template comment at the top of every one of those files lists exactly `DiagnosticType, Severity, Path, Moniker, Justification` — so the near-universal element was the one being dropped, and the three rare ones were the ones always emitted. `<Justification>` is now always written; when the caller supplies none it gets an obvious `TODO:` plus a warning, rather than being invented or silently omitted.

### 3. Added `path` as the preferred input

Sub-elements — controls, fields, methods, enum values — **cannot** be addressed by type+name at all. Real paths go to 22 segments (`dynamics://Enum/SRSReportPrintOrientation/EnumValue/Potrait?Value`). The finding hands you the path anyway, which is precisely what MS's template says: *"(path given in warning messsage)"*. `elementType`+`elementName` stays as the fallback for top-level elements.

Also: `AxEdt` isn't a real element type (it's `AxEdtString` / `AxEdtInt` / `AxEdtInt64` / `AxEdtEnum` / `AxEdtDate` / `AxEdtGuid` / `AxEdtReal`), and `AxQuery` is `AxQuerySimple`. The union is now the measured set — a test asserts every element type present in real files is in it.

### 4. `validate` gave a bare ✅ to strings that aren't BP rules

168 of the 577 entries are `canonical: false` — they came only from the rule-DLL resource dump, which also carries **upgrade- and form-conversion-tool** messages: `DEC*` (~20), `ActionPane*` (10), `CopiedStandardProperties`, `ChangedIgnoreEDTRelationToYes`. Checked against a real `AxRuleSet/BPRules.xml`:

```
DECReplaceCode                    in BPRules.xml: False   -> tool said "✅ is a real BP moniker"
ActionPaneUpdateActionPaneStyle   in BPRules.xml: False   -> tool said "✅ is a real BP moniker"
BPErrorPrivilegeNotCoveredByDuty  in BPRules.xml: True    ✓
```

`canonical` is the only field that answers "is this a BP rule". A resource-only hit now gets a qualified answer instead of a ✅.

### 5. The stated coverage limitation was backwards

The PR body, the module docblock, the generated header and the script all say most X++-authored rules carry no message/description. The generated data says the opposite: **545 of 577 entries carry a real message (94%)**, 221 also carry a description, and only **32** are name-only.

This one had teeth, because it reached the model: the zero-result search message told the caller *"most X++-authored rules have no message/description text ... so a miss here does not mean no rule fits"* — i.e. discount a negative result that is in fact reasonably reliable. In a server where a round trip costs real money, teaching the agent to distrust a good negative is a live cost. Corrected in all four places.

### 6. Smaller ones

- **`nearMisses` was unreachable.** It filtered for entries whose lowercase equals the input's lowercase — but `BY_MONIKER` is keyed lowercase, so such an entry would already have hit and `entry` would be non-null. Always `[]`, and the "Found with different casing only:" branch in `bpMonikerHelp.ts` was dead. Replaced with word-overlap `suggestions`, which addresses the miss that actually happens (`BPErrorPrivilegeNotCoveredByDuties` → suggests `BPErrorPrivilegeNotCoveredByDuty`). `found` still stays `false` — a candidate to confirm, never a correction applied silently.
- **Multi-placeholder templates.** `.replace(/\{\d+\}/g, elementName)` filled `{0}`, `{1}` and `{2}` all with the same name: `"Number of Lines found in class method MyClass is greater than MyClass within class MyClass"`. That's an invented sentence in `<Message>`. Now only a single-placeholder template is filled; otherwise `<Message>` is omitted (absent from 57% of real entries anyway).
- **`parseBpFindings` misread a real sample in this repo.** `tests/tools/runBpCheck.test.ts:465` and `:907` contain `BPError: LocalVariableNotUsed` — `BPError` is a severity prefix, not a rule name. It parsed as `moniker: "BPError"` and stamped *"⚠️ not in the extracted moniker catalog — verify the spelling"* onto output the compiler had just emitted. That's the most expensive kind of false alarm: it invites a round trip to re-verify something already authoritative.

### Schema budget

I did **not** raise the ceiling again. Only `action`/`moniker`/`path`/`justification` are published on the wire; `elementType`, `elementName`, `limit`, `message`, `severity`, `itemSpecific` stay in the zod handler schema — the same trade this file already makes for `labels`. Payload measures 51,642 against the 51,700 you'd already set.

### Verification

- `tsc --noEmit` clean
- **3,732 tests passed, 0 failed** across 299 files (full suite)
- 418 real VM entries replayed through the generator, `<Path>` matching byte for byte

### Still open

The catalog itself is untouched — still from 10.0.2527.174 on your box. Regenerating against a different install would tie it to that version and produce a large diff, so I left that call to you.

PR #908, opened 2026-08-12
fix(form-extension): place an added control by resolving its parent

kklyandev:
## Issue

`add-control` on a form extension inserted the new control with
`content.replace('</Controls>', …)`. A string pattern replaces the **first**
occurrence, so when the extension defines its own container the nested
`</Controls>` — which closes earlier in document order — received the control.
`parentControl` was never resolved to a node at all; it only supplied the
`<Parent>` text.

Two defects follow from that one line:

1. The control was wrapped in an `AxFormExtensionControl` envelope regardless of
   where the parent lived, so a parent the extension itself defines got the
   base-form representation. A form extension has two mutually exclusive shapes —
   an envelope in the root `<Controls>` for a base-form parent, a bare
   `<AxFormControl>` in the parent's nested `<Controls>` for an extension-owned
   one — and only the parent's location distinguishes them.
2. With two sibling groups in one file, a second call inserted into the **first**
   call's parent while writing the requested name into `<Parent>`.

**Neither fails the build.** The malformed file compiles at exit 0, 0 errors: the
deserializer discards the misplaced node, and the control is absent from compiled
metadata and from the running form. On a `DataGroup`-bound parent the sole trace
is a metadata warning naming the field group — never the control, the node, or
the tool. On a plain group the compile logs are byte-identical with and without
the control, so there is no diagnostic anywhere in the toolchain.

## Fix

Placement now comes from where the resolved parent lives: an
`AxFormExtensionControl` envelope in the extension's root `<Controls>` for a
base-form parent, a bare `<AxFormControl>` in the parent's nested `<Controls>`
for one the extension defines. The file is walked by offset rather than
re-serialized, so untouched bytes stay untouched.

Because the compiler will not catch this for us, the write is validated before it
lands. `findFormExtensionPlacementProblems` asserts that every
`AxFormExtensionControl` is a direct child of the root collection and that every
nested collection holds only `AxFormControl`. It gates both the writer's own
output and hand-written `xmlContent`, where the existing name-based checks passed
the malformed shape clean — every element in it is spelled correctly; only its
position is wrong.

Also on this path:

- The success banner reported `applied via IMetadataProvider.Update()` for writes
  the XML fallback performed, naming an API that never ran — and with it, implying
  a validation that never happened. It now names the path that did the write.
- The control-type auto-pick answered `ComboBox` for every enum. `NoYes` and
  `NoYesId` now resolve to `CheckBox`, matching what the designer emits, via the
  EDT chain so a custom EDT extending `NoYesId` resolves too.
- `previousSibling` positions the control instead of being silently dropped, and
  says so when it cannot apply.
- Idempotency keys on resolved control names instead of a `<Name>` substring
  search, which also matched data sources, fields and the extension's own name —
  turning a real add into a silent no-op.
- An extension-owned parent carrying `<DataGroup>` now warns rather than refuses.
  The base-form guard refuses because a base-form `<DataGroup>` container
  generates its members and an explicit control collides; an extension-created
  group generates nothing — a field-group member with no explicit control does not
  render — so there the explicit control is the only thing that puts the field on
  the form.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,26 @@
+# Line endings are decided here rather than by each contributor's core.autocrlf.
+#
+# Without this file the answer came from whatever the local Git setting happened
+# to be, and an editor that rewrote a file's endings turned a 50-line change into
+# an 800-line diff — the content edit becomes unreviewable, and `git diff` needs
+# --ignore-cr-at-eol to say anything useful. Everything here is text and normal
+# to read as a diff, so everything is LF, in the repository and in the working
+# tree, on every platform.
+
+* text=auto eol=lf
+
+# Shell scripts break outright with CRLF; stated explicitly so the rule survives
+# any future narrowing of the wildcard above.
+*.sh text eol=lf
+
+# Binary: never normalise, never diff as text.
+*.png binary
+*.jpg binary
+*.jpeg binary
+*.gif binary
+*.ico binary
+*.pdf binary
+*.zip binary
+*.dll binary
+*.exe binary
+*.snk binary
--- a/src/tools/specs/d365foFileOpSpecs.ts
+++ b/src/tools/specs/d365foFileOpSpecs.ts
@@ -165,11 +165,14 @@ export const D365FO_FILE_PARAM_SPECS: Record<string, { type: string; description
   controlLabel: { type: 'string', description: 'Optional label for the new control.' },
   positionType: {
     type: 'string',
-    description: 'AfterItem | BeforeItem. Omit to append at the end of the parent.',
+    description:
+      'AfterItem (needs previousSibling) | Begin | End. Omit to append at the end of the parent. ' +
+      'These are the values D365FO form-extension metadata actually carries; anything else is refused.',
   },
   previousSibling: {
     type: 'string',
-    description: 'Name of the sibling control to position after (used with positionType=AfterItem).',
+    description:
+      'Name of the sibling control to position after. Implies positionType=AfterItem when that is omitted.',
   },
   baseFormName: {
     type: 'string',
@@ -436,10 +439,15 @@ export const D365FO_FILE_OP_SPECS: Record<string, D365FileOpSpec> = {
       + 'Otherwise pass the exact name of an existing container (Tab, TabPage, Group, Grid). '
       + 'objectType="form-extension": parentControl may name a base-form container OR a container '
       + 'the extension itself defines — the writer picks the right XML shape from which it is, so '
-      + 'just pass the name. A parent bound to a table field group via <DataGroup> is REFUSED: its '
-      + 'children mirror that field group, so add the field to the field group instead '
-      + '(add-field-to-field-group) and refresh the group in the designer. previousSibling positions '
-      + 'the control only under an extension-defined parent.',
+      + 'just pass the name. A BASE-FORM parent bound to a table field group via <DataGroup> is '
+      + 'REFUSED: the compiler generates that group\'s members, so an explicit control collides — add '
+      + 'the field to the field group instead (add-field-to-field-group) and refresh the group in the '
+      + 'designer. On an EXTENSION-OWNED <DataGroup> parent the control IS written, with a warning: '
+      + 'nothing tops that group up, so the explicit control is what puts the field on the form, and '
+      + 'the field group entry is needed as well so a later designer Refresh does not discard it. '
+      + 'previousSibling works under either parent — under an extension-defined parent it orders the '
+      + 'control in the XML, under a base-form parent it is written as '
+      + '<PositionType>AfterItem</PositionType> + <PreviousSibling>.',
   },
   'add-enum-value': {
     required: ['enumValueName'],
--- a/src/tools/specs/d365foFileOpSpecs.ts
+++ b/src/tools/specs/d365foFileOpSpecs.ts
@@ -433,7 +433,13 @@ export const D365FO_FILE_OP_SPECS: Record<string, D365FileOpSpec> = {
     ],
     note: 'objectType="form": parentControl="Design" adds the control at the TOP LEVEL of the '
       + 'form design — use it for the first control on a form whose design is still empty. '
-      + 'Otherwise pass the exact name of an existing container (Tab, TabPage, Group, Grid).',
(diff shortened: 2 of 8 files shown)


dynamics365ninja:
Reviewed this and pushed three follow-up commits to the branch. The core of the PR is right — resolving `parentControl` against the extension's own control tree and deriving both the shape and the offset from where it lives is the correct fix, and the placement validator is well grounded. Everything below is on the edges, and most of it lands on one specific person: the user whose files the *shipped* writer already damaged, who runs this release against them.

## What changed

### Repair paths (`92d4e17`)

**The post-write check gated on zero placement problems in the whole document.** Any pre-existing misplacement aborted an otherwise-correct write with *"this is a bug in the writer, not in your call"* — so the release that fixes the writer was also the release that locked every already-damaged file out of the repair, while blaming the caller. It now diffs against the file's own baseline, by problem **count** rather than by line (insertion shifts every line below it), and refuses only what the write introduced. Pre-existing damage comes back as a note.

**Idempotency and parent resolution were position-blind.** A control inside a misplaced element is discarded whole by the deserializer, but its `<Name>` reads back perfectly. So re-running the identical `add-control` — the one repair a user would naturally try — answered `✅ already present … skipped (idempotent)` and wrote nothing; and naming such a control as `parentControl` nested the new one inside the dead subtree, discarded along with it. Controls are now split into live and discarded, and only live ones count as present or usable as a parent.

**A childless extension-owned parent was refused.** That broke this tool's own two-step workflow: `add-control` can create a group (`group → AxFormGroupControl`), and the group it writes has no `<Controls>` at all, since `innerControlLines` emits only `Name`/`Type`/`FormControlExtension`. So every "create a group, then add a child to it" sequence failed on step two with *"Add the first child in the form designer"* — sending the caller to Visual Studio for the exact job the tool exists to do. It now creates the collection. The position isn't a guess: across the 1088 shipped `AxFormExtension` files, an opening `<Controls>` follows `FormControlExtension` 2176 times and `ControlModifications` 996 times, and nothing else, ever.

**`previousSibling` was discarded for a base-form parent**, with the note *"Order within the extension's root `<Controls>` does not affect layout."* That part is true, but the conclusion isn't: position among the base form's children is carried by the envelope itself. Shipped metadata is explicit — 753 files use `Name, FormControl, Parent, PositionType, PreviousSibling`, e.g. `ApplicationSuite/Foundation/AxFormExtension/BusinessProcessActionLookup.Foundation.xml` positions `RetailTaskActionType` after `TaskActionType` in base-form group `View`. Now emitted, which also gives `positionType` its first reader — it was declared in the op spec *and* the zod schema and read nowhere in `src/`. Accepted values are `AfterItem` and `Begin` (765 and 182 occurrences; the only two in the corpus), `End` means omitting both, and anything else is refused rather than guessed at — an unknown enum value deserializes to a discarded node with the build reporting 0 errors, which is the failure mode this module exists to prevent. The spec's `BeforeItem` does not appear in shipped metadata.

**The op spec contradicted the runtime.** It advertised a blanket refusal for a `<DataGroup>`-bound parent, while the code writes the control with a warning for an extension-owned one — and that warning offered the field group as a substitute one bullet before stating it would not work. Since the model reads the spec before calling, it would pick the action the runtime documents as ineffective. Both now say the same thing: on an extension-owned parent, both halves are required.

### Error attribution (`72334be`)

The placement banner passed the **control** name into a parameter declared `objectName`, so it read `form-extension "NewCtl" …` and never named the file — in the report-a-bug path, where identifying the object is the whole job.

A refusal from our own XML writer was surfaced as `{ success: false }` and rendered by the shared bridge path as `Bridge operation 'add-control' failed: …`, naming an API that never ran. That's the same defect `viaXmlFallback` was introduced to fix on the success branch. It also fed `isUnresolvedObjectError`, so a deterministic file-based refusal could buy a provider refresh and a full retry to reach the identical answer. Both branches now check the marker that was already being set — and this turned out to be wider than `add-control`: `modify-property`'s *"appears N times — ambiguous"* and `add-index`'s *"has no fields"* were being reported as bridge failures too.

One correction to my own review note: no refusal currently in the tree actually trips `isUnresolvedObjectError`. That was latent risk, not a live bug; it's now closed structurally.

Also adds `.gitattributes` (`* text=auto eol=lf`). There was none, so line endings came from each contributor's `core.autocrlf` — an editor rewriting one file's endings turned a 307-line change into an 834-line one that needed `--ignore-cr-at-eol` to read. Committed blobs are already LF, so `git add --renormalize .` reports no content change.

### Escaping (`0c9afe8`)

Found while testing the above. `controlName`, `typeValue`, `dataField`, `dataSource`, `label`, `wrapperName`, `parentControl`, `previousSibling` and the `i:type` attribute were all interpolated raw. Object names can't carry markup, but **labels are free text**: `Cost & freight` emitted a bare ampersand — not well-formed, and rejected by the deserializer. A value carrying markup could go further and close its own element, smuggling a complete `<AxFormExtensionControl>` into the parent's nested `<Controls>`, which is exactly the shape that gets silently discarded. The post-write guard caught that, but not causing the damage beats detecting it.

Escaping on write forces decoding on read, or the round trip breaks: a control written as `Weird &amp; Name` would never match `Weird & Name` on the next call, and the idempotency check would add it twice. `textValueOf` decodes; every comparison and every displayed name goes through it.

## Verification

- `tsc --noEmit` clean; **300 test files / 3762 tests pass** (+30 from this branch's follow-ups)
- The writer run over **all 1088 shipped `AxFormExtension` files** in `PackagesLocalDirectory`: 1088 writes, **0 refusals, 0 introduced placement problems** — unchanged before and after escaping, which is the property that matters there
- Every fix that could be pinned both ways was: reverting it makes its test fail. Two that couldn't are called out below

## Known gaps

- The abandoned-write path is unreachable from ordinary input — the writer's output is clean across the whole shipped corpus, and escaping closed the one route that reached it. Its message builder is exported (`buildAbandonedWriteMessage`) and tested directly instead. It takes the document and derives the object name itself rather than accepting a pre-picked string, so the call site has nothing left to get wrong.
- Three existing tests asserted the old behaviour (the `<Controls>` refusal, the `previousSibling` "ignored" note, and the contradictory `DataGroup` bullet). They were rewritten rather than deleted.

PR #913, opened 2026-08-13
feat(bp-moniker): regenerate each instance's catalog from its own D365FO version

laeliand:
Addressing the concern about moniker catalog versioning. 

catalog.generated.ts is one compiled-in snapshot shared by every instance, regardless of which D365FO version that instance is actually pinned to (Scenario F runs several instances against different versions at once).

Adds a dist single mode and per-instance catalog instead: extract-bp-catalog.ps1 gains a JSON output mode (-OutFile/-Version), bpMonikers/index.ts loads it via a new BP_CATALOG_PATH setting when present and falls back to the compiled default otherwise, and ensureBpCatalogFresh() wires it into rebuildIndex() so it regenerates automatically on instance add/rebuild/upgrade (and the first-time setup wizard) only when the instance's resolved version has actually moved — every other call is a no-op, and a missing pwsh/powershell or a failed extraction just logs a warning rather than failing the reindex.

Keep shipping the default DevBox catalog as cheap insurance, not redundant if catalog cant be generated locally.


dynamics365ninja:
## Review — findings and fixes

Reviewed the branch and pushed the fixes as `152ebbd` + `66561a5`. Everything below is verified rather than inferred: the two headline findings were reproduced on a Windows Server 2022 box running Windows PowerShell 5.1 against a real `K:\AosService\PackagesLocalDirectory`.

The design is right — one compiled snapshot shared across instances pinned to different versions is a real problem, and `ensureBpCatalogFresh` hangs off exactly the choke point that already sees creation, upgrade, rebuild and the wizard. The findings are all about the feature being a **silent no-op on the boxes it targets**: nothing errored, nothing logged, it just never did anything.

### 1. `extract-bp-catalog.ps1` could not be parsed by Windows PowerShell 5.1 at all

The file is BOM-less UTF-8 (as it should be — `.gitattributes` makes everything LF text). PowerShell 5.1 reads a BOM-less file as the ANSI codepage, so a `U+2014` em dash (`E2 80 94`) arrives as three CP1252 characters ending in `U+201D` — a smart quote, **which PowerShell accepts as a string delimiter**. So this line:

```powershell
throw "No PackagesLocalDirectory found under $dynamicsDir — pass -PackagesPath explicitly."
```

closed its string early and the parse died with `Unexpected token 'pass'` — before a single statement ran. The CLI falls back to `powershell` whenever `pwsh` is missing, and on a stock D365FO dev box `where pwsh` finds nothing, so the fallback path *is* the normal path and could never have worked.

This one predates the PR (it reproduces on `main`, line 63) — but this PR is what starts invoking the script automatically, which turns it from latent into load-bearing. The script is now pure ASCII, like its six siblings.

> Verification trap worth knowing: PowerShell's `>` and `Out-File -Encoding utf8` **add** a BOM, so `git show HEAD:script.ps1 > copy.ps1` in a PowerShell session silently "fixes" the file and makes a before/after parse comparison meaningless. My first measurement showed a false regression for exactly this reason. Extract exact bytes outside PowerShell, then parse with `[System.Management.Automation.Language.Parser]::ParseFile`.

### 2. `Set-Content -Encoding utf8` writes a BOM under 5.1, and `JSON.parse` rejects it

BOM-less under pwsh 7, BOM-prefixed (`EF BB BF`) under 5.1. Both readers then failed, and both failed *silently*:

- `loadCatalog()` caught the throw and fell back to the compiled snapshot — the whole feature becomes a no-op;
- `existingVersionKey()` also threw → returned `null` → the version never matched, so the full recursive `Get-ChildItem` over the entire PackagesLocalDirectory re-ran on **every** `instance rebuild`/`upgrade`, which is precisely the "every other call is a cheap no-op" property the design rests on. (Measured on this box: 13.8 s and a 25-assembly reflection pass per rebuild, against 0.0 s for the intended no-op. Not catastrophic — but it is pure waste, and it multiplies by instance count.)

Now written with `[System.IO.File]::WriteAllText(..., New-Object System.Text.UTF8Encoding($false))`, with the relative-path case resolved against `(Get-Location).ProviderPath` first (.NET resolves relative paths against the process CWD, not the PS location). Both readers also strip a leading BOM, for catalogs an older copy of the script already produced.

**End-to-end after the fix**, under 5.1 against the real packages root: exit 0, 577 entries, first byte `0x7B`, Node parses it with no BOM handling, and the version stamp round-trips — so the second call is finally the promised no-op. Through `loadCatalog` that gives 577 entries, 545 with rule text, 409 canonical.

### 3. A non-empty `default` on `index.bpCatalogPath` made `BP_CATALOG_PATH` always set

`defaultPathEnv()` (`src/config/configFile.ts`) projects **every** `type: 'path'` setting that has a string default onto `process.env`. With the default present, `if (!overridePath) return COMPILED_CATALOG` in `loadCatalog()` was unreachable, and every install that has never regenerated a catalog — all existing ones, every Linux/Azure deployment, every dev checkout — hit `readFileSync` ENOENT and printed a warning on **each process start**. It also contradicted the line this PR adds to `docs/CONFIGURATION.md` ("only written once an instance has regenerated its own catalog").

The default is gone; `ensureBpCatalogFresh` writes the value after a successful extraction, which is what the description already promised. A test pins the property, since it looks exactly like something worth "tidying up" later.

### 4. The written value was the resolved **absolute** path

That breaks the portable-relative-path invariant instance settings rely on (`configBaseDir`, `src/utils/loadEnv.ts`): rename `instances/clientA` → `clientB` and `BP_CATALOG_PATH` still points into the old folder → ENOENT warning plus a full re-extraction. Now writes `./data/bp-moniker-catalog.json`; `toEnvRecord()` resolves it against `baseDir` on the way out anyway.

### 5. `frameworkDirectory ?? ''` turned "unknown packages root" into "let the script guess"

`listXppConfigs()` deliberately keeps configs whose JSON could not be read, so `frameworkDirectory` can legitimately be `undefined`. Passing `''` isn't "use the default" — the script's own `if (-not $PackagesPath)` branch auto-detects the **newest** `%LOCALAPPDATA%\...\PackagesLocalDirectory` on the box, and the result then gets stamped with *this* target's version and treated as current. A catalog from the wrong install is worse than no catalog. Returns `null` now.

### 6. `resolvePinnedXppConfig` fell back to the newest config when the pin didn't resolve

Which is not what its docstring claimed, and not what `XppConfigProvider.getActiveConfig` does — that returns `null` for a named-but-absent config and only auto-selects when no name was given (it also matches the short name, which the new helper didn't). Concretely: an instance pinned to `contoso___10.0.2428.63` after a UDE upgrade removed that config — exactly the state `isXppConfigStale()` flags and `instance upgrade` exists to fix — would silently get its catalog extracted from the 10.0.2500 FrameworkDirectory and stamped as current, while its index and server still referenced the stale pin. Now mirrors `getActiveConfig` exactly.

### 7. The override was accepted with no validation beyond `Array.isArray`

`entries: []` passed the check and permanently replaced a 577-entry catalog with nothing: every `validate` misses, `search` returns nothing, and `buildSuppressionXml` warns "not in the extracted catalog (0 known monikers)" — the opposite of the module's stated guarantee. That's reachable from a *successful* run (a packages root with `bin/` but no `AxRuleSet/BPRules.xml` and no BP rule DLLs exits 0 with zero entries).

Separately, `BP_MONIKER_CATALOG.map(e => e.moniker.toLowerCase())` runs **outside** the `try`, so any entry lacking a string `moniker` threw at module import and took the server down with it — against the docblock's "a bad override must never make `bp_moniker` unusable".

Empty is now treated as a failed extraction, and malformed entries are dropped with a warning.

### 8. Hardcoded `545` mixed with a dynamic catalog length

`545 of the ${BP_MONIKER_CATALOG.length} entries carry real rule text` was true for the one compiled snapshot; per-instance it can render as "545 of the 402 entries". Now computed (`MONIKERS_WITH_RULE_TEXT` — which does evaluate to 545 on the measured install, so the literal was right, just no longer safe). The version-specific 545/577/221/32 figures in both docblocks are attributed to the install they were measured on rather than stated as fact.

### Smaller things

- `.DESCRIPTION` still said "Output is a generated TypeScript module, not JSON" while `.PARAMETER OutFile` twenty lines below documented the JSON mode this PR adds.
- `bpMonikers/index.ts` pointed at `indexCmd.ts` for `ensureBpCatalogFresh`; it's defined in `bpCatalog.ts`, which `indexCmd` only calls.
- `rebuild-instance.ps1` held the last non-ASCII byte in the repo's scripts. It's in a comment so it parsed fine, but it's the same latent trap — removed, so the guard can cover all seven scripts instead of just the one that happened to break.

### Checked and found fine

Module import order (`bootstrapEnv.js` runs before `bpMonikers` is evaluated, so `loadEnv()` has already populated the env); `paths.extractBpCatalogScript` resolving for both a checkout and an npm install, with the script correctly added to `package.json` `files`; `-Version`/`-OutFile` passed after `-File` so they bind to the script rather than the host; `commandExists('powershell')` exiting rather than hanging; the PS `if`-as-hashtable-value syntax on 5.1; the traditional-environment branch correctly returning `null` when `bin\` is absent; and `Get-Help` still parsing the comment-based help after the docblock edits.

### Tests

+29 across three new files (`powershellScriptEncoding`, `xppConfigPin`, `bpCatalogPathDefault`) and two extended. The PowerShell guards are source-text assertions on purpose — CI runs on Linux and cannot reproduce either 5.1 behaviour, so the script text is the only place these regressions can be caught. They assert the rules (ASCII-only, BOM-less, no `Set-Content` for the JSON payload) rather than the specific symptoms, with a vacuity check so an empty `readdir` can't make the whole file pass silently.

Full suite green: 3898 passed / 2 skipped, typecheck and lint clean, `docs/CONFIGURATION.md` regenerated so `config-docs-up-to-date` passes.

### End-to-end verification on a real instance

Unit tests can only simulate the two links that matter most here — the setting write, and the server actually picking the catalog up — so I drove the real `ensureBpCatalogFresh` against a real `instanceTarget()` on a traditional environment (`K:\AosService\PackagesLocalDirectory`), then threw the instance away:

| | |
|---|---|
| first run | 13.8 s, 577 entries, first byte `0x7B` (no BOM), 409 canonical / 545 with rule text |
| persisted setting | `"index.bpCatalogPath": "./data/bp-moniker-catalog.json"` — relative, and merged alongside the existing `environment` block rather than replacing it |
| second run | **0.0 s, "BP catalog up to date"** — the no-op the design promises, working for the first time |
| `loadEnv` | resolves `BP_CATALOG_PATH` to the catalog inside that instance's folder |
| `bpMonikers` | loads 577 entries from the instance, not the compiled snapshot; `MONIKERS_WITH_RULE_TEXT` = 545 (so the literal that was hardcoded happened to be right for this version — just no longer safe to hardcode) |
| after renaming the instance folder | `BP_CATALOG_PATH` follows it and the catalog still loads — finding 4's portable-path fix, confirmed rather than argued |

**One gap I could not close:** the UDE path has never been exercised live. This box is traditional-only, so `resolvePinnedXppConfig` → `frameworkDirectory` → extraction from a UDE `FrameworkDirectory` is covered by unit tests with fixtures only. That is the scenario the PR exists for, and the specific unknown is whether a UDE `FrameworkDirectory` really is the root that has `bin/BPExtensions/*.dll` and `AxRuleSet/BPRules.xml` beneath it — the layout differs from a traditional VM. If it isn't, extraction exits 0 with zero entries; that now falls back to the compiled catalog instead of silently emptying `bp_moniker` (finding 7), but the feature would do nothing on UDE. **Worth one `d365fo-mcp index <instance>` on a UDE box before merging** — a non-zero entry count matching that instance's version is the whole test.

PR #919, opened 2026-08-16
fix(build): point -compilermetadata at the model store, not the framework dir

kklyandev:
## Scope

This applies to **UDE only**. A UDE box keeps custom source in a model store
(`ModelStoreFolder`) that is a different directory from Microsoft's binaries
(`FrameworkDirectory`). On CHE the two are the same `PackagesLocalDirectory`, so the
problem below cannot occur there and the fix is a no-op.

## Issue

xppc writes its compiler metadata back to the `-compilermetadata` root during its
"Metadata Write-Back" phase. The build tool aimed that at the `FrameworkDirectory`, so
every UDE build deposited `<FrameworkDirectory>\<Model>\XppMetadata`.

Two consequences:

- **Custom model names accumulate in a shared directory.** Nothing cleans them up, and
  each stub is indistinguishable from a real package to anything enumerating that
  folder — it carries only an `XppMetadata` child, no `Descriptor`, no `bin`, no source
  XML.
- **The `-incremental` baseline is split.** Visual Studio writes its compiler metadata
  to the model store; this tool wrote to the `FrameworkDirectory` and copied forward
  into the model store via `syncCompilerMetadata`, with nothing copying back. A build
  here following a VS build compared against metadata that predated it. `-incremental`
  is the default path, so this is the common case.

## Fix

Point `-compilermetadata` at the model store — what VS does. The `FrameworkDirectory`
keeps its other roles (locating `xppc.exe` and `labelc.exe`, the Dynamics DLLs, and
`-referenceFolder`), and Microsoft's compiler metadata still resolves through the
reference folder.

`generateRuntimeMetadata` now takes the compiler-metadata root as a parameter,
defaulting to `microsoftPackagesPath` for existing callers, so `syncCompilerMetadata`
looks where xppc actually wrote. With both roots equal it becomes a no-op reporting
"already co-located" rather than a misleading "no compiler metadata to sync".

Additionally, the xppc invocation is now recorded at the top of the build log.
`buildLog()` sends it to stderr and to `bridgeLogFile` (which requires
`D365FO_BRIDGE_LOG_FILE`), so the build log carried no record of the arguments that
produced it. That part applies to every environment.

### DIFF
--- a/src/tools/sdlc/buildProject.ts
+++ b/src/tools/sdlc/buildProject.ts
@@ -1,7 +1,7 @@
 import { execFile, spawn } from 'child_process';
 import util from 'util';
 import path from 'path';
-import { access, writeFile, readFile, unlink, appendFile, readdir, stat } from 'fs/promises';
+import { access, writeFile, readFile, unlink, appendFile, readdir, rm, stat } from 'fs/promises';
 import { openSync as openSyncFs, closeSync as closeSyncFs } from 'fs';
 import os from 'os';
 import crypto from 'crypto';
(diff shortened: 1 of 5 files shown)


dynamics365ninja:
Audited this and pushed one follow-up commit (edcb613) to the branch — the core change is sound, but it left `run_bp_check` aiming at the old location.

### `run_bp_check` was still pointed at the framework directory

`src/tools/sdlc/runBpCheck.ts` passed `-compilerMetadata=<microsoftPackagesPath>`. Before this PR every build deposited `<FrameworkDirectory>\<Model>\XppMetadata` there, so xppbp saw a fresh baseline; after it, nothing writes there again. On a UDE box that means xppbp is handed a root that no longer contains the module's compiler metadata.

I checked what xppbp actually does in that state rather than assuming, against **xppbp 7.0.7996.33**, simulating the UDE split with a metadata root holding a single module and the framework packages root alongside it:

| `-compilerMetadata` root | result |
|---|---|
| holds the module's `XppMetadata` | `Errors: 0`, rules evaluated normally |
| does not hold it (state after this PR) | the checked class itself reported *"appears not to have been compiled"*, then every element of the module, `CompilerMetadataMissing`, exit `-1` |

So the rules that read compiled X++ silently do not run, and what comes back reads like a clean object.

The fix follows from xppbp's own help text — these are two flags, not two spellings of one:

```
-packagesRoot(pr)=path       The path to the packages root containing binaries for modules.
-compilerMetadata(cm)=path   The path to the compiler metadata.
```

Both are now passed: `-compilerMetadata` = model store (where this PR makes xppc write it), `-packagesRoot` = framework directory (binaries of referenced Microsoft modules). Confirmed in the table above that this combination resolves the framework side fine. `compileModelLabels` was sharing the same variable while genuinely needing the framework directory for `labelc.exe`, so it now takes that under its own name.

`describeNonRun()` also recognises the `CompilerMetadataMissing` warning **for the requested element**, so a run whose rules never looked at the X++ can no longer be reported as findings-free. Deliberately scoped to the named target: a whole-model run routinely carries this warning for extension classes whose base class the environment does not have installed.

### Three smaller ones from the same audit

- **Stale stubs are never cleaned up.** Every build made before this change left `<FrameworkDirectory>\<CustomModel>\XppMetadata` behind, and the framework directory is still passed as `-referenceFolder` (line 598 loop) — so xppc keeps finding metadata frozen at the last pre-move build, which is the classic *"has not been successfully compiled since it was last changed"* shape. Added a narrow cleanup: UDE only, only for a model that lives in the model store, and only when the folder holds nothing but `XppMetadata` (a write-back stub, not a package deployed there on purpose). Anything else is logged and left alone; a build is never failed over it.

- **The new invocation header never reached a failed build's response.** `readFullLog` only returns head+tail when *no* diagnostic line matches; as soon as one is present — the exact case someone audits — it returns diagnostic windows plus the last 30 lines. For any log over 300 lines the header added to answer "which root did `-compilermetadata` point at" was in the file but never in the output. The header markers are now shared with the reader, which always keeps those lines. The test asserts the diagnostic-window path specifically, so it cannot pass vacuously through the head+tail fallback.

- **`buildLog` hardcoded `customPackagesPath`** instead of the value actually used (`ctx.compilerMetadataPath`, set 14 lines later). They agree today only by coincidence — now read from `ctx`.

### Note, not a change

The `-packagesRoot` fallback styles (older xppbp with no `-compilerMetadata`) still point at the framework directory, as before. One flag cannot carry both roots on a split-root box, and no version old enough to need that path has been observed on UDE — commented in place rather than changed on speculation.

Full suite green on the updated branch: 312 files, 3936 passed (+13), `tsc --noEmit` clean.

Issue #825, opened 2026-08-07
Tool schemas are 61 KB re-sent on every request, and the 105-tool catalogue forces the host into tool_search mode


dynamics365ninja:
Measured in the session audit (#824). The tool catalogue sent to the model is 150 KB across 105 tools, of which this server contributes 26 tools / 61 335 B. Two tools account for a quarter of our surface:

| Tool | Total | description | JSON schema |
|---|---|---|---|
| `d365fo_file` | 9 771 B | 818 B | **8 855 B** |
| `generate_object` | 8 488 B | 835 B | **7 545 B** |
| `labels` | 5 857 B | 822 B | 4 946 B |
| `search` | 4 345 B | 552 B | 3 703 B |
| `object_patterns` | 3 950 B | 919 B | 2 933 B |
| `extension_info` | 2 863 B | 749 B | 2 017 B |
| `prepare` | 2 180 B | 480 B | 1 613 B |

Two separate problems fall out of this.

## 1. The schemas are re-sent on every request

Tool definitions were ~15,3k tokens of every one of the 89 requests — 37,4 AIU together with the system prompt, **12 % of the session**, for bytes that never change. `d365fo_file` and `generate_object` alone are 16,4 KB of it.

Both are big because they inline a discriminated union of every operation and its parameters into one JSON Schema. That is the right ergonomics for a human reading the schema and the wrong shape for a prefix that ships 89 times.

Suggestion: keep `operation` (or `action`) as a closed enum with one-line-per-operation descriptions, and demote `params` to a loose object. Move the per-operation parameter contract into `get_knowledge` / `d365fo_file_op_specs` (which already exists as `d365foFileOpSpecs.ts`) so the agent fetches it *once, for the operation it actually picked*, instead of carrying all of them all the time. Same for `generate_object`.

Rough arithmetic: taking those two tools from 16,4 KB to ~4 KB saves ~3,3k tokens per request ≈ **6 AIU per 19-minute session**, and it scales linearly with session length.

## 2. The catalogue size changes host behaviour

At 105 tools VS Code stops sending the catalogue inline and switches to a search-based tool surface: the model must call `tool_search` to discover tools before it can use them. In this session that cost **6 dedicated round trips**, each one a solo turn at ~3 AIU, ~18 AIU + ~20 s of pure discovery overhead — and the searches returned only 5 hits apiece, so the model went back repeatedly:

```
[...]

PR #922, opened 2026-08-18
feat(d365fo_file): add remove-control, remove-entry-point and a delete action

laeliand:
Three removal capabilities the tool was missing, each the inverse of something it could already do. Without them the only way to take a control, a privilege entry point or a whole object out of a model was
d365fo_file(action="create", overwrite=true) — the whole-file rewrite that loses every collection the caller did not think to re-type.

remove-control (form, form-extension)
  Walks the nested <Controls> tree by resolved control name, so no parentControl
  is needed and a data source sharing the control's name is not mistaken for it.
  On a form-extension the whole <AxFormExtensionControl> envelope goes, not just
  its <FormControl> — an envelope without its control is a <Parent> reference to
  nothing. removeSeparator (opt-in) drops the orphaned AxFormButtonSeparator
  sibling and says that it did. An emptied collection collapses to <Controls />.

remove-entry-point (security-privilege)
  Removes one <AxSecurityEntryPointReference>, by entryPointName or by
  entryPointObjectName (+ entryPointObjectType). Two matches are REFUSED rather
  than resolved: removing the wrong entry point revokes access to a different
  object, builds clean, and surfaces only as a user losing a form. Lives next to
  buildAxSecurityPrivilegeXml so the two halves of the shape cannot drift, and
  round-trips against it byte-for-byte.

action="delete"
  Undoes both halves of a create: the XML, and the <Content Include> in EVERY
  .rnrproj of the model that lists the object — cleaning only the active project
  is exactly the case that leaves a dangling include and a project that will not
  load. Un-registers before unlinking so a locked project file is caught while
  the object is still whole, then clears the symbols, labels and undo-ledger
  entry that would otherwise outlive it. An object that resolves to nothing is
  an error, not a silent no-op: "already gone" and "the name was wrong" look
  identical to the caller, and reporting the second as success leaves the object
  in the build under a ✅. Standard-model, cross-model and path-containment
  refusals are the same ones every other write clears.

Neither remove has a C# op behind it (MetadataWriteService exposes no RemoveControl, and security objects have no bridge write path at all), so both are direct-XML writers admitted through canBridgeModify via a per-operation XML_ONLY_MODIFY_PAIRS map. Dropping security-privilege into BRIDGE_MODIFY_TYPES instead would have claimed add-method/replace-code/modify-property work on a privilege, and sent the caller after a bridge resolution failure that cannot apply.

The offset-preserving XML tag reader moves out of formExtensionControlXml.ts into utils/xmlNodeTree.ts, since the removal writer needs the same tree over an AxForm's <Design>. A second copy that has to get CDATA, comments and quoted attributes right is the kind of duplication that drifts silently.

Ratchets updated deliberately: 31 -> 33 operations, d365fo_file schema cap 5,300 -> 5,700 chars (measured 5,567), total 51,700 -> 52,100, and the operation count in docs/MCP_TOOLS.md. Parameters stay behind get_knowledge(kind="op-spec") per #825; only the enum values and one line for the new action are published.


dynamics365ninja:
Audited this branch. The design holds up — the per-operation `XML_ONLY_MODIFY_PAIRS` gate rather than widening `BRIDGE_MODIFY_TYPES`, refusing two entry-point matches instead of resolving them, and returning a refusal rather than `null` from the XML writers so the caller is not sent after a bridge-resolution cause that cannot apply, are all the right calls. `xmlNodeTree.ts` is a byte-for-byte extraction.

Three findings are the silent kind, though — each produces a state the caller is told is fine — plus a few smaller ones. Fixes are up as #923, cut from this branch, so **this PR merges first** and #923 then shrinks to its own two commits.

**1. `delete` leaves a dangling `<Content Include>` and reports the opposite.** `resolveMembership` decides ownership through `includeKey` (lowercased, `.xml` tolerated); `removeFromProject` compares the raw attribute with `===`. A project whose include differs only in case is named as an owner, then matches nothing on removal — so the file is unlinked, the include stays, and the reply says *"No project of model X referenced `AxForm\Name` — nothing to un-register."* Reproduced against real `.rnrproj` content:

| include in the project | membership | removal | include survived |
|---|---|---|---|
| `AxForm\ConDemoTicketTable` | active | `true` | no |
| `AxForm\condemoticketTABLE` | **active** | **`false`** | **yes** |
| `AxForm\ConDemoTicketTable.xml` | **active** | **`false`** | **yes** |

Case drift is documented in `projectMembership.ts`'s own header ("…CtsoFinExtension" on disk against "…CtsoFINExtension" in the XML). `deleteD365File.test.ts` cannot see it because `mockRemoveFromProject` returns `true` unconditionally.

**2. The `<Controls>` collapse drops the namespace reset.** `formControlRemoval.ts` writes a literal `'<Controls />'`, discarding the open tag's attributes. Shipped AxForm metadata declares `xmlns="Microsoft.Dynamics.AX.Metadata.V6"` on the root and resets it on every element under `<Design>` — a real collection is `<Controls xmlns="">`, and the bare form puts the element back in the V6 namespace. Checked against `ApplicationSuite\Foundation\AxForm`. `insertFormExtensionControl` preserves exactly these attributes in the other direction. Form extensions are unaffected; their root `<Controls>` genuinely has no `xmlns`.

**3. `delete` skips `enforceGrounding`,** which `create` and `modify` both apply to `*-extension` objects. Under `GROUNDING_ENFORCE=true` an agent barred from *creating* a table extension without a prepare token is free to *delete* one.

Smaller ones, all fixed in #923: `objectType` is not checked against the file's own `Ax<Type>` folder (and `axFolderForObjectType` falls back to `'AxClass'`, so a mismatched type un-registers a different object); the cross-model refusal says "Refusing to modify" and offers an extension as the remedy, which cannot un-define anything; `remove-control` can match a discarded twin over the live control on an already-damaged file, where `insertFormExtensionControl` distinguishes the two; "there is no undo for it" is pessimistic — `undo_last_modification` restores a git-tracked XML, just not the project entries.

#923 also carries one pre-existing defect the audit surfaced, unrelated to this PR: `addToProject` has the same divergence in the other direction and writes a *duplicate* `<Content Include>` for an entry spelled differently. `normalizeInclude` is now the single definition for all of it.

Verification on #923: `tsc` clean, 315 files / 4009 tests green, and each of the 11 new tests was run against the reverted source to confirm it actually fails without the fix.

PR #926, opened 2026-08-20
fix(bp-check): revert -compilerMetadata to framework dir to fix BPFrameworkFatalException

laeliand:
edcb613 moved compilerMetadataPath from the framework directory to the model store to fix CompilerMetadataMissing (PR #919 moved xppc's XppMetadata write-back to the model store). That fix introduced a BPFrameworkFatalException crash on every run: xppbp's StaticRuntimeProvider constructs {compilerMetadataPath}\StaticMetadata\AxView.md, which does not exist in the source model store — those compiled .md files are deployment artifacts that only live in the framework packages directory.

Revert compilerMetadataPath back to frameworkPath (microsoftPackagesPath). CompilerMetadataMissing returns for compiled-X++ rules on UDE, but a warning is far better than a fatal crash that prevents BP checks from running at all. describeNonRun() already gives clear guidance in that case.

### DIFF
--- a/src/tools/sdlc/runBpCheck.ts
+++ b/src/tools/sdlc/runBpCheck.ts
@@ -512,6 +512,33 @@ export const runBpCheckTool = async (params: any, context: any) => {
     const frameworkPath = microsoftPackagesPath || packagesRoot;
     const compilerMetadataPath = frameworkPath;
 
+    // On UDE, xppbp's DataEntityDuplicateLabelChecker constructs a RuntimeMetadataProvider
+    // using the -metadata path as runtimeRoot and then looks for
+    // {runtimeRoot}\StaticMetadata\AxView.md. That file is a deployment artifact that only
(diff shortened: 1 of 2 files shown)


laeliand:
I need to revaluate the issue again

laeliand:
On UDE, ModelStoreFolder (source XML) and FrameworkDirectory (deployed binaries + StaticMetadata\) are two separate roots. xppbp's -metadata flag was designed for CHE where they're the same directory — there's no second flag to split the two concerns.

It's unclear how Visual Studio itself handles this with the parameters xppbp exposes. Rather than guess or copy files, the junction creates a filesystem-level pointer: {ModelStoreFolder}\StaticMetadata → {FrameworkDirectory}\StaticMetadata. From xppbp's perspective, everything it needs is under one root — identical to what CHE provides natively — with no data duplication.

PR #933, opened 2026-08-24
fix(enums): kernel enums were reported as hallucinated symbols

dynamics365ninja:
## The defect

`NoYes` is defined by the X++ **runtime**, not by an AOT element. There is no `AxEnum/NoYes.xml` in any package — verified against the complete set of **8,220 `AxEnum/*.xml` basenames** under `K:\AosService\PackagesLocalDirectory` (metamodel 7.0.7996.33) — so all three resolution paths correctly fail to find one:

- the C# bridge: `IMetadataProvider.Enums` does not carry it (a `search` for "NoYes" with `limit: 100` returns 21 enums *containing* the string — `NoYesBlank`, `NoYesCombo`, `DefaultNoYes` — and never `NoYes` itself);
- the SQLite symbol index: it indexes AOT XML, and a direct query returns **0 rows** for `NoYes` while `NoYesBlank` is present, so the index is healthy;
- a disk probe: there is no file.

Every path then treated *absent from the AOT* as *does not exist*. Two consequences, both reproduced live on this VM.

### 1. The validator called the most common enum in D365FO a hallucination

```
validate_code(mode="references", codeType="xml-table")  on  <EnumType>NoYes</EnumType>

❌ 1 issue(s) found (1 error(s), 0 warning(s))
❌ <EnumType>NoYes</EnumType>
   Enum "NoYes" not found in the symbol index.
Fix errors before writing — these will cause compiler failures or wrong object references.
```

`NoYes` appears **48 times in `CustTable.xml` alone**, and EDT `NoYesId` resolves to it. So the validator was instructing agents to edit correct metadata, with a hard error and a compiler-failure warning. `search` then offers `NoYesBlank` / `NoYesCombo` / `DefaultNoYes` as the plausible "corrections" — and those enums are **real**, so the resulting edit compiles clean, passes BP, and means something else entirely. That is the silent-wrong-write shape, produced by the tool whose job is to prevent it.

### 2. The reader sent callers to index a file that cannot exist

```
get_object_info(objectType="enum", name="NoYes")

Enum "NoYes" not found via bridge, symbol index, or on disk.
1. `search` for `NoYes` — the exact name may differ (model prefix, casing, suffix).
2. Real object in a custom package that isn't indexed yet? Run update_symbol_index({ filePath: "<absolute path to NoYes.xml>" }) …
3. Just created it this session? …
```

None of that can ever succeed. And the server hands the agent the name in the first place — `get_object_info(edt, NoYesId)` reports `Enum Type | NoYes` — so the loop has no exit. `Exception` and `Types` behave identically.

## Root cause

The X++ reference resolver **already knew this**. `resolveReferences.ts` allow-listed `'noyes'` and `'exception'`, with a comment saying why:

> Kernel enums (not in metadata XML). 'exception' has no AxEnum at all, and the index does not prove 'noyes' — without both, every try/catch and NoYes:: use hard-errors in the static-access path.

Which is why X++ using `NoYes::Yes` validated cleanly (`✅ all 6 reference(s) verified`) while the XML path hard-errored on the same enum in the same session. **The knowledge existed in exactly one path.**

## The fix

`src/knowledge/kernelEnums.ts` holds it once, and the three paths that need it use it:

- **`validateCode.ts`** — a kernel enum counts as verified for `<EnumType>`, instead of erroring.
- **`enumInfo.ts`** — answers authoritatively up front: what it is, that there is nothing to index, that `<EnumType>` and `NoYes::Value` are valid, and specifically **not** to substitute `NoYesBlank`/`NoYesCombo`/`DefaultNoYes`.
- **`resolveReferences.ts`** — sources its kernel-*enum* half from the shared module rather than spelling it out, so the two lists cannot drift apart again.

**Membership is evidence-based, not remembered.** All 12 names were checked against the 8,220 AOT enum basenames and are absent from every one: `NoYes`, `Exception`, `Types`, `TableScope`, `ConcurrencyModel`, `StatementType`, `IsolationLevel`, `UtcDateTimeOrder`, `DateOrder`, `DateDay`, `DateMonth`, `DateYear`.

Value names come from a scan of shipped X++ under PackagesLocalDirectory and are labelled **observed rather than exhaustive** — the kernel is the authority. Where the scan found no usage (`DateOrder`, `IsolationLevel`, …) the values are **omitted rather than guessed**, and a test enforces that.

## Live verification

Against the real index (`data/xpp-metadata.db`, 2.5 GB) through the real code path, bridge absent — the same state the reader hits when `IMetadataProvider` has no such enum:

| Probe | Result |
|---|---|
| `NoYes` rows in the real index | `[]` — and `NoYesBlank` present, so the index is fine |
| `<EnumType>NoYes</EnumType>` | ✅ "all 2 reference(s) verified" — **was a hard ERROR** |
| `<EnumType>FmTotallyInventedEnum</EnumType>` | ❌ still an error — the check keeps its teeth |
| `get_object_info(enum, NoYes)` | "# Enum: NoYes (kernel enum)" — **was not-found + go-index-it** |
| `get_object_info(enum, FmNoSuchEnumAnywhere)` | still reports not-found |

The scratch harness used for this is not committed (it needs a 2.5 GB local DB); the committed tests cover the same logic with a mock faithful to `lookupSymbolsNocase`'s two-probe query shape, including the case that proves the validator did **not** start accepting everything.

## Tests

`327 test files, 4804 passed.` `tsc --noEmit` clean, `biome lint` clean on the touched files. New:

- `tests/knowledge/kernelEnums.test.ts` — membership, case-insensitivity, the guarantee that real AOT enums whose names merely *contain* a kernel one (`NoYesBlank`, `DefaultNoYes`, `HcmBlankNoYes`) are **not** swallowed, and that the description omits values rather than inventing them.
- `tests/tools/xmlEnumKernelRefs.test.ts` — the regression itself, plus "still catches an enum that really does not exist".

### DIFF
--- a/src/knowledge/kernelEnums.ts
+++ b/src/knowledge/kernelEnums.ts
@@ -1,0 +1,131 @@
+/**
+ * Kernel enums — enums the X++ runtime defines, with NO AOT metadata element.
+ *
+ * These are referenced by ordinary metadata (`<EnumType>NoYes</EnumType>` appears
+ * 48 times in CustTable.xml alone, and EDT NoYesId reports `Enum Type: NoYes`)
+ * but they have no `AxEnum/*.xml` anywhere, so no metadata-backed lookup can
+ * ever prove them: not the C# bridge (`IMetadataProvider.Enums` does not carry
+ * them), not the SQLite symbol index (it indexes AOT XML), not a disk probe.
+ *
+ * Treating "absent from the AOT" as "does not exist" produced two failures, both
+ * confirmed live on 2026-08-24:
+ *
+ *   1. `validate_code(mode="references", codeType="xml-table")` reported
+ *      `<EnumType>NoYes</EnumType>` as a hard ERROR — "Enum NoYes not found in
+ *      the symbol index" — under "Fix errors before writing — these will cause
+ *      compiler failures". An agent obeying that edits correct metadata, and
+ *      `search` offers it NoYesBlank / DefaultNoYes / NoYesCombo to "correct" to.
+ *      Those enums are real, so the result compiles clean and means the wrong
+ *      thing.
+ *   2. `get_object_info(objectType="enum", name="NoYes")` answered "not found via
+ *      bridge, symbol index, or on disk" and advised searching other spellings,
+ *      running update_symbol_index on `NoYes.xml`, and checking
+ *      D365FO_CUSTOM_PACKAGES_PATH. None of that can ever succeed: there is no
+ *      file to index. Meanwhile the EDT reader hands the agent the name `NoYes`
+ *      in the first place — a loop with no exit.
+ *
+ * The X++ reference resolver already knew this (resolveReferences.ts allow-listed
+ * 'noyes' and 'exception' for exactly this reason), which is why X++ using
+ * `NoYes::Yes` validated cleanly while the XML path hard-errored on the same
+ * enum. This module is that knowledge in one place, so the next path to need it
+ * does not have to rediscover it.
+ *
+ * MEMBERSHIP IS EVIDENCE-BASED, not remembered: every name below was checked
+ * against the complete set of 8,220 `AxEnum/*.xml` basenames in
+ * K:\AosService\PackagesLocalDirectory (metamodel 7.0.7996.33) and is absent
+ * from all of them.
+ */
+
+/** A kernel enum, and the value names it is used with. */
+export interface KernelEnum {
+  /** Canonical casing, for rendering. */
+  name: string;
+  /**
+   * Value names as OBSERVED in shipped X++/metadata under
+   * PackagesLocalDirectory. Deliberately not claimed to be exhaustive — it is
+   * what the product itself uses. Omitted where a scan found no usage, rather
+   * than filled in from memory.
+   */
+  values?: string[];
+}
+
+const ENTRIES: KernelEnum[] = [
+  // By far the most common: boolean-ish fields reach it through EDT NoYesId.
+  { name: 'NoYes', values: ['No', 'Yes'] },
+  { name: 'Exception', values: [
+    'Break', 'CLRError', 'CodeAccessSecurity', 'DDEerror', 'Deadlock',
+    'DuplicateKeyException', 'DuplicateKeyExceptionNotRecovered', 'Error',
+    'FunctionArgument', 'Info', 'Internal', 'Sequence', 'Timeout',
+    'TransientSqlConnectionError', 'UpdateConflict',
+    'UpdateConflictNotRecovered', 'ViewDataSourceValidation', 'Warning',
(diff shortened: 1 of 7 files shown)


dynamics365ninja:
## Follow-up: the allow-list alone was not enough

A full scan of every `<EnumType>` reference in PackagesLocalDirectory against the 8,220 AxEnum basenames finished after the first commit, and it changes the shape of this fix.

**44 enum names that shipped Microsoft metadata references cannot be resolved by the symbol index.** Measured through the real 2.5 GB index using the same `lookupSymbolNocase` the checker calls, so casing is not the explanation. Among them:

`TableGroup`, `AccessRight`, `AccessType`, `AccessLevel`, `SortOrder`, `JoinMode`, `MenuItemType`, `MessageSeverity`, `OpenMode`, `ClientType`, `AOSClientMode`, `RangeStatus`, `UtilElementType`, `UtilEntryLevel`, `UtilFileType`, `Units`, `AggregateFunction`, `FormTreeCheckedState`, `DatabaseLogType`, `SecurableType`, …

`TableGroup` alone is a property of essentially every table.

So the hard error was never really about `NoYes` — that was the loudest instance of a class. Extending the allow-list name by name would have been guesswork dressed as evidence: I can prove these 44 are referenced by shipped metadata and unresolvable, but not that each one is a *kernel* enum rather than one from a module this installation lacks.

### What changed

Unresolvable enums are now a **warning**, not an error. The check is index-only, it cannot tell "absent from the index" from "does not exist", and it was asserting the second under *"these will cause compiler failures"*. An agent obeying that picks a real-but-different enum out of search results, and the edit compiles clean meaning something else — the exact outcome the check exists to prevent. The `<Extends>` check in the same function already warns on this reasoning.

The verified kernel names stay silent via `kernelEnums.ts`, so the common cases cost nothing.

### Two tests were passing for the wrong reason

- `resolve-references.test.ts` seeds `sym('NoYes', 'enum')` into its mock index, so *"accepts a known enum (NoYes) in `<EnumType>`"* proved only that the mock was more generous than any real index. It now runs against an index that proves nothing — what a real one actually offers for `NoYes`. **The X++ side had already spotted this trap** (*"The shared `deps` seeds NoYes, which would mask the fix — prove the allow-list"*) and tested against `emptyDeps`; the XML side had not. That is how the hard error shipped and survived.
- *"flags an unknown enum"* asserted `isError`; it now asserts the finding is reported as a warning, with the reason.

### Live, against the real index

| `<EnumType>` | in index | result |
|---|---|---|
| `NoYes`, `Exception` | no | ✅ verified silently (kernel list) |
| `TableGroup`, `AccessRight`, `SortOrder` | no | ⚠️ warning, call succeeds |
| `FmInventedXyz` | no | ⚠️ warning, still reported |

`327 test files, 4805 passed.`

PR #941, opened 2026-08-24
fix: the two loose ends from the audit

dynamics365ninja:
These were the two items #939's description listed as *"not fixed here"*. I fixed them and pushed to that branch — but the merge had already happened, so the commit never reached `main`. This is that commit, cherry-picked onto current `main`.

## 1. A green build could swallow a warning

`trimSucceededLog` kept only lines matching `DIAG_LINE_TEST`, which is anchored and case-sensitive — it wants `[Kind ]Error: ` / `[Kind ]Warning: ` at the start of the line. Measured against the real matcher, it **keeps**

```
Metadata Warning: dynamics://…       Compile Error: Class dynamics://…
BEST PRACTICE Warning: …             Warning: something generic
```

and **drops**

```
warning: lowercase generic
MyTable.xpp(12,3): warning CS1234: unused variable
```

On the **failure** path that is harmless — a non-matching line still arrives through the head/tail fallback. On the **success** path it is dropped outright, and `hasWarnings` uses the same test, so such a warning would not even set the ⚠️ icon. Invisible twice over. Confirmed by running it: an MSBuild-shaped warning did not survive a green build.

The costs are not symmetric. A handful of extra lines is nothing; a silently dropped warning is the one thing this function must not do. Anything that **mentions** an error or a warning is now kept too, whatever its shape.

## 2. `search` was missing from `PLURAL_FORM_TOOLS`

It has `queries[]` (max 10), and the batch covers the tool's primary use — exactly the criterion the list's own comment states for inclusion. Its absence meant the session analyzer scored **zero** avoidable round trips for repeated single searches, which is the most common repeat pattern there is.

The audited fixture is unchanged: that session had no consecutive `search` turns, so the watch is added without inventing a finding.

## Not changed

A tail with no trailing summary still keeps 12 arbitrary rows under an accurate `[N phase-timing line(s) omitted]` label. Detecting "summary" would be guesswork and the cost is twelve lines.

`331 test files, 4831 passed.`

### DIFF
--- a/src/cli/session/analyze.ts
+++ b/src/cli/session/analyze.ts
@@ -52,6 +52,7 @@ const PLURAL_FORM_TOOLS = [
   'get_object_info', 'run_bp_check', 'verify_d365fo_project',
   'd365fo_file',    // operations[] applies many edits to one object in a call
   'get_knowledge',  // topics[] answers several lookups in a call
+  'search',         // queries[] runs several searches in one call, in parallel
 ];
 
 export interface CostRates {
--- a/src/tools/sdlc/buildProject.ts
+++ b/src/tools/sdlc/buildProject.ts
@@ -416,7 +416,7 @@ export function trimSucceededLog(logTail: string, keepTail = 12): string {
   const diagnostics: string[] = [];
   let omitted = 0;
   for (let i = 0; i < summaryFrom; i++) {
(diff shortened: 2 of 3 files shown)


dynamics365ninja:
Superseded — reopening from fix/audit-findings instead, so the whole audit lives on one branch.

PR #943, opened 2026-08-25
fix(config): stop silently pinning the first project in multi-project models

laeliand:
Summary
configManager's D365FO_SOLUTIONS_PATH fallback used to pin whichever .rnrproj scanned first when several VS projects shared the target model — writes then silently landed in an arbitrary project the caller never asked for. It now mirrors the existing ambiguousProjects pattern elsewhere in the file: resolve the model, but leave the project unset, requiring an explicit projectName/projectPath to pick one.

Reworded the projectName/projectPath descriptions on get_workspace_info and d365fo_file so agents pass them proactively whenever they already know the target project (a plan, ticket, or prior context), not only on an explicit user "switch" — while preserving the existing guardrail that these params only steer where writes land, never a way to read another model's data (reads already span every model).

Added a regression test (tests/utils/solutionsPathAmbiguity.test.ts) covering the multi-project-same-model case, plus a control case for ordinary single-project resolution.

Why

A workspace with many .rnrproj files under one model would silently resolve to the alphabetically/scan-order-first project on every fresh session, regardless of which project the caller was actually working in. The tool's own docstring discouraged passing projectName/projectPath unless a human literally said "switch project," so calling agents rarely overrode it even when they knew the correct target from context.


dynamics365ninja:
Audited this branch and pushed two follow-up commits (`5a241ec`, `186781c`) plus a merge of current `main`. The direction here is right and the regression test is a real one — it fails against the pre-fix code and the control case passes. What I found was that the refusal, once it fires, left the caller with nothing to act on.

Worth stating up front how often it fires: a real solution on our VM has **190 `.rnrproj` across 31 models, the largest model built by 20 of them**. `sameModelProjects.length > 1` is the ordinary shape, not the edge case, so everything below is on the common path.

### What was fixed

**The alternatives were recorded and then never shown.** Nothing read `autoDetectedProject.ambiguousProjects`. The create warning's candidate list comes from `getWorkspaceProjectCandidates()`, which only the WORKSPACE scan fills — and it is empty on exactly this route, because the solutions-path scan only runs when the workspace held no `.rnrproj` at all. So the caller got the bare three-line "no projectPath could be resolved" with no names in it, and `get_workspace_info` said `(not detected)` — a broken-configuration message for a state where nothing is broken and a choice is missing. Added `getAmbiguousProjectPaths()`; both sites now name the projects, and `get_workspace_info` reports `Project : (not selected — N projects build this model)` followed by their file names, so the agent does not have to spend a second call on `diagnostics=true` to learn what to pass.

**`recordDetectionSuccess` still reported the project that was deliberately not selected.** Line 328 kept `primary.projectPath ?? null`, so `d365fo-mcp doctor` printed a projectPath next to a `get_workspace_info` that says none was chosen — two answers for one state. The sibling call at line 256 already passes `detectedProject.projectPath ?? null`, which is correctly null for ambiguous results; this one now matches.

**The field copying was inverted.** The new object copied `packagePath`/`packageName`, which `scanAllD365Projects` never sets (it returns only `projectPath`/`modelName`/`solutionPath`), and dropped `solutionPath`, the one field it always sets. `getSolutionPath()` returning null here also silently killed the `solutionPath → ProjectFileFinder.findProjectInSolution` fallback in `createLabel.ts:802` and `createD365File.ts:4493`. Now `...primary` with `projectPath` explicitly dropped, and `solutionPath` survives when every candidate agrees on it (a real fact about the model then) and goes when they disagree (as much a guess as `projectPath` — handing it out just moves the arbitrary pick into `findProjectInSolution`).

**The guard was anchored on `primary`, which is itself whichever project scanned first.** With several models under one root, whether the caller is protected is decided by disk order: if the first-scanned project happens to own a unique model, the `else` branch pins it. I did *not* make this a refusal — for the many workspaces whose only model source is this scan, withholding the model would break detection outright, and the project choice does follow coherently from the resolved model. But the log now says the model **was a pick, by scan order**, and names what else was under the root.

**`get_workspace_info` answered a refused `projectName` with nothing else.** It is the first call of every session and the parameter is one this PR now tells the agent to pass from context, so a miss is expected traffic — and it cost the whole call plus a round trip to ask again without the argument. The refusal now comes first (still `isError`, so a rejected switch cannot read as a completed one), followed by the workspace facts the call was made for.

**Two smaller ones.** `Projects : N in solution (N-1 besides the active one)` subtracted an active project that may not exist — now `(none selected)`. And `docs/CONFIGURATION.md` still described the scan as "what makes multi-project workspaces work without reconfiguring"; the note about the model-vs-project split went into `src/config/settings.ts` (the generated file is checked by `config-docs-up-to-date`), plus a CHANGELOG entry.

### On the schema wording

Kept the change, tightened it. "You already know it from context — a plan, a ticket" invites exactly what the next sentence forbids: a plan or ticket names a *model* or a feature, not a `.rnrproj` stem, and a model selects no project. It now reads "pass it whenever you have one: the USER named it, or an earlier call listed it. NOT a model name, and not a guess from a ticket". The phrases `toolInventory.test.ts` pins are preserved. `projectPath` gets its `D365FO_SOLUTIONS_PATH` mention back — with it unset, `getAllDetectedProjects()` is empty and `projectName` cannot resolve at all, so `projectPath` is the only route. The total ListTools payload is back under the 45 000-char ratchet (I trimmed a redundant sentence in the top-level description to pay for it).

### Test changes

Six new regressions across `solutionsPathAmbiguity.test.ts` and a new `workspaceInfoProjectAmbiguity.test.ts`, each confirmed to fail against the code before these commits, with the control cases passing. The `workspaceDetector`/`fs-promises` mocks now spread the real module — the hand-listed version silently dropped every export added later (`distinctCustomModels` was one, and it surfaced as "no export is defined on the mock" from three unrelated tests).

Full suite green: 352 files / 5024 tests, typecheck and biome clean, all 7 CI checks passing.

PR #949, opened 2026-08-27
fix(bridge): fill in an EDT's inherited StringSize instead of the type default

kklyandev:
## Issue

`get_object_info(objectType="edt", name="ItemFreeTxt")` reports `StringSize` **10**. The real size is **1000**, inherited from `ItemFreeTxtBase`.

`IMetadataProvider` returns an EDT exactly as its own XML declares it and does not fill in what it inherits. When a derived string EDT declares no `StringSize`, the instance reports the `AxEdtString` constructor default of 10 rather than the inherited size. `ItemFreeTxt` reads back as 10 and is really 1000; `ItemId` and `CustAccount` read back as 10 and are really 20.

The table reader carried the same defect at `MapField`, and is the higher-traffic consumer: across ten core tables, **310 of 564 string fields (55%)** reported the wrong size. `InventTable.ItemId` said 10 against a real 20; `InventTable.AltConfigId` said 10 against a real 50.

Two rules constrain any fix, both measured against 10.0.2645 rather than assumed:

- **A derived EDT can declare its own `StringSize`**, and 228 in the shipped corpus do (`InvoiceId` declares 50 over `Num`'s 20). The platform permits it when the base carries `StringSizeIsExtensible = Yes`. A declared value is authoritative and must survive.
- **A declared value does not always win.** Over the 182 derived EDTs that declare a size and have a declaring ancestor: with `StringSizeIsExtensible = Yes` the declaration wins in either direction (`CustInvoiceId` keeps 20 under `InvoiceId`'s 50); otherwise a declared value *smaller* than the nearest declaring ancestor's loses to it (4 EDTs — `PartyName` declares 100 under `DirPartyName`'s 160 and is 160, plus `ITMJourneyId`, `TAMRebateInvoice`, `PSNPurchasingCardProviderName`). No `Auto`-and-larger EDT ships, so that direction is unverified and the fix assumes nothing about it.

The value an undeclared child takes comes from the **nearest** ancestor that declares one, not the root: `CorrectedInvoiceId_RU` → `InvoiceId(50)` → `Num(20)` resolves to 50.

## Fix

Defer to the metamodel's own resolvers rather than reimplementing the rules above:

- `ReadEdt` calls `MetaEdtExtensions.ResolveVirtualProperties`
- `MapField` calls `MetaTableFieldExtensions.GetStringSize`

Both are passed `isFlightEnabled: false`, which is not a detail. With `true`, an EDT-level read of `CustInvoiceId` returns 50 while the field-level resolver returns 20 for `CustInvoiceJour.InvoiceId` — a field typed with that very EDT — so the bridge would contradict itself between `get_object_info(edt)` and `get_object_info(table)`. With `false` the two agree at 20. The flag changes nothing for any of the broken cases (`ItemFreeTxt` 1000, `ItemId` 20, `AccountNumber_IN` 20 under either value).

Each has a hand-rolled fallback, because the bridge is compiled against one environment's Microsoft assemblies and loaded against another's, where the helper may be absent. The fallback fills in **only** an undeclared size, from the nearest declaring ancestor — that rule matched the resolver on 148 of 148 affected EDTs. It deliberately does not reproduce the override rule for a declared-but-smaller value: leaving a declared number alone is the honest failure mode, and that case covers 4 EDTs corpus-wide. The field-level fallback latches off after the first failure so a missing helper cannot emit one warning per string field per table read.

Mutating the returned instance is safe — the `DiskProvider` deserializes a fresh object on every `Read`, verified by reference-comparing two reads, so nothing leaks into a later read or into the write path.

`NoOfDecimals` is left alone throughout. It looks similar but is not virtual: a child real EDT may declare its own (`AccruedAmountMST_IT` declares 2 while its base reports -1), so copying an ancestor's would overwrite a real local value.

Also fixed on the way: the "not set" sentinel for `DisplayLength` is **-1, not 0**, so the old `== 0` test never fired and every EDT without a local `DisplayLength` (24,446 of 25,332) reported -1.

Reporting changes:

- The EDT report names where the number came from — `| String Size | 1000 (inherited from ItemFreeTxtBase) |` — and says nothing when the number is the EDT's own. A `Root EDT` row is added for orientation in multi-level hierarchies.
- The SQLite fallback used to omit `String Size` entirely for derived EDTs (12,354 of 25,332 indexed EDTs declare none) and now walks `edt_metadata.extends` to the nearest declaring ancestor. It cannot see `StringSizeIsExtensible`, so the override rule is documented as a known limitation of that path.

Every figure above comes from the metamodel resolvers against 10.0.2645 — the same source the compiler uses, but not an independent oracle. None of it was cross-checked against live SQL column widths.

### DIFF
--- a/bridge/D365MetadataBridge/Models/Models.cs
+++ b/bridge/D365MetadataBridge/Models/Models.cs
@@ -1,4 +1,4 @@
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Text.Json.Serialization;
 
 namespace D365MetadataBridge.Models
@@ -255,6 +255,14 @@ namespace D365MetadataBridge.Models
         [JsonPropertyName("extends")]
         public string? Extends { get; set; }
 
+        /// <summary>
+        /// Root of the Extends chain, or null for a root EDT. Reported for orientation in a
+        /// multi-level hierarchy; it is not necessarily the element that declares the
+        /// StringSize -- see StringSizeInheritedFrom for that.
+        /// </summary>
+        [JsonPropertyName("rootEdt")]
+        public string? RootEdt { get; set; }
+
         [JsonPropertyName("label")]
         public string? Label { get; set; }
 
@@ -263,6 +271,14 @@ namespace D365MetadataBridge.Models
 
         [JsonPropertyName("stringSize")]
         public int? StringSize { get; set; }
+
+        /// <summary>
+        /// Ancestor the reported StringSize came from. Null when the number is what this EDT's
+        /// own XML declares. Set both when the EDT declares nothing (the common case) and when
+        /// an ancestor's value overrode one the EDT did declare.
+        /// </summary>
+        [JsonPropertyName("stringSizeInheritedFrom")]
+        public string? StringSizeInheritedFrom { get; set; }
 
         [JsonPropertyName("referenceTable")]
         public string? ReferenceTable { get; set; }
--- a/bridge/D365MetadataBridge/Models/Models.cs
+++ b/bridge/D365MetadataBridge/Models/Models.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Text.Json.Serialization;
 
 namespace D365MetadataBridge.Models
@@ -256,9 +256,13 @@ namespace D365MetadataBridge.Models
         public string? Extends { get; set; }
 
         /// <summary>
-        /// Root of the Extends chain, or null for a root EDT. Reported for orientation in a
-        /// multi-level hierarchy; it is not necessarily the element that declares the
-        /// StringSize -- see StringSizeInheritedFrom for that.
+        /// Root of the Extends chain. Reported for orientation in a multi-level hierarchy; it is
+        /// not necessarily the element that declares the StringSize -- see StringSizeInheritedFrom
+        /// for that.
+        ///
+        /// Null for a root EDT, and also null when the chain could not be walked to its end -- a
+        /// dangling Extends or a cycle -- because the last ancestor reached is not the root and
+        /// naming it as one would be a wrong answer rather than a missing one.
         /// </summary>
         [JsonPropertyName("rootEdt")]
         public string? RootEdt { get; set; }
--- a/bridge/D365MetadataBridge/Services/MetadataReadService.cs
+++ b/bridge/D365MetadataBridge/Services/MetadataReadService.cs
@@ -438,7 +438,7 @@ namespace D365MetadataBridge.Services
 
             try { var mi = prov.Tables.GetModelInfo(tableName); if (mi?.Count > 0) result.Model = mi.First().Name; } catch { }
 
-            try { foreach (var f in table.Fields) result.Fields.Add(MapField(f)); } catch (Exception ex) { Warn("fields", tableName, ex); }
+            try { foreach (var f in table.Fields) result.Fields.Add(MapField(f, prov)); } catch (Exception ex) { Warn("fields", tableName, ex); }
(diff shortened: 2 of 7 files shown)


dynamics365ninja:
Reviewed this and pushed a follow-up commit onto the branch (c4edc63) rather than leaving a list — the C# half needed a rebuild against the metadata assemblies to prove out, and that only happens on a D365FO machine. Everything below is about how the code behaves off the platform it was measured on; the defect you found is real and the resolution rules are right.

**The two hand-rolled fallbacks could not run for the failure they exist for.** Both resolvers were called directly with the `try`/`catch` inside the method holding the call. The CLR resolves a method token when it JITs the method *containing* the call, so `MissingMethodException` surfaces at that method's own call site, one frame up, and the inner catch never fires. Measured on Framework64 v4.0.30319 against an assembly with the helper removed — the inner catch did not run and the fallback line never printed. `Program.cs:24` already documents this exact hazard and works around it with `[MethodImpl(NoInlining)]`.

Consequences were worse than the wrong number: out of `ResolveInheritedEdtProperties` it escapes into `ReadEdt`, which has no guard, and out of `EffectiveFieldStringSize` it escapes into `MapField` → the table's `try { foreach … } catch { Warn("fields", …) }`, so the table returns with its **field list truncated at the first string field**, reported as a success with one stderr line. The latch meant to bound that noise never executed either.

The compile side is the bigger one: `bin/` is gitignored and excluded from `package.json` `files`, the csproj sweeps drive letters for PackagesLocalDirectory/bin, and `bridgeClient.ts:1023` tells the user to `dotnet build` — the bridge is compiled on each user's own machine. On a platform without the helper a direct call is `CS0117`, so that user gets **no bridge at all**, where today they get one reporting 10. Both helpers are now bound by name through reflection (the precedent is already in this file, at the patterns-DLL load), so absence becomes `helper == null` → the fallback, announced once, and the project still compiles.

**Also found while fixing it:** `FindInheritedStringSize` called `PickProvider` unguarded, and `PickProvider` *throws* by design when a provider fails — deliberately, so a single-kind lookup can never report "does not exist" for "could not look". Here it decorates a read that already succeeded, so that throw could fail a good EDT read, or a whole table's field list, over an ancestor nobody asked about.

The rest, briefly:

- **`RootEdt` could be wrong rather than missing.** It was the last ancestor *reached*; a chain broken by a dangling `Extends` or a cycle ends at something that is not the root. Verified live: `AccruedAmountMST_IT → AmountMST → MoneyMST` (no AxEdt file) now reports no root instead of `AmountMST`. Tracked as `ChainComplete`.
- **`StringSizeInheritedFrom` took its number from Microsoft's resolver and its attribution from our own walk**, with nothing checking the two agree, so the label could credit an ancestor declaring something else. Dropped when they disagree; the number stands.
- **The field-level latch** turned any exception into a permanent process-wide downgrade. Absence is handled by the binding now, so what reaches the catch is a per-field failure (a dangling `ExtendedDataType`) — it warns once and keeps using the resolver.
- **Chain-walk cost.** `readEdt` is a loop caller (`createD365File` issues one per distinct EDT of a new table) and every hop is a full XML deserialize; hops now read from the child's own provider first and only pay the two-provider `Exists` probe on a miss.
- **The 10-is-unset argument** was measured for EDTs and only assumed for fields — said so where the assumption lives.
- **Node side:** both SQLite modes walked the same chain with two separate loops and the same two stop conditions, and hierarchy mode showed only what each level *declares*, i.e. nothing for the EDT the caller asked about. They now share `walkEdtChain`, and hierarchy mode states the effective size once. That walk also stops applying `modelName` past the start row — a child in an ISV model extending a Microsoft one is the normal case, and carrying the filter up cut the chain at the first hop.
- Dropped the UTF-8 BOM `Models.cs` picked up, which lands in the attested `sourceHash`.

**Verified:** `npm run bridge:build` clean (0 warnings), attestation regenerated. The rebuilt bridge run against 10.0.2645 reproduces every case in your description — ItemFreeTxt 1000 from ItemFreeTxtBase, ItemId 20 from ItemIdBase, AccountNumber_IN 20 from ExternalAccount (not CustVendAC), PartyName 160 from DirPartyName, InvoiceId 50 with no inheritance claimed, InventTable 145 fields with ItemId 20 and AltConfigId 50 — with no warnings on stderr, so the reflection binding resolves both helpers on a current platform. 5039 tests pass; CI green.

Happy to split any of this back out if you'd rather land it separately.

PR #961, opened 2026-08-30
Post-Phase-F: report scaffold fixes, golden hygiene, five language cases captured (core coverage 100%)

dynamics365ninja:
Closes the three items Phase F (#959) found on the VM and left open in the report
generator, tidies what that change owns in the golden catalog, and authors the five
eval cases the closure queue had been asking for.

## Generator (evidence: `eval/corpus/runs/2026-08-30T04+`)

- **EDT suggestion is model-aware.** `resolveBestEdt` takes an optional context: a
  field of that name already in the target model wins outright, and a candidate that
  is another model's prefix glued onto the field name is demoted below the name
  heuristic. That is what made a bare `fieldsHint` pick `PlCorrNoteId` (un-migrated,
  `BPErrorEDTNotMigrated`), `smmSubject` and `ReqGroupId`. Opt-in, so callers that
  name no model behave exactly as before. Demoting foreign prefixes exposed a second
  hole — `LineCount` would have fallen to the string default — so `*Count` now
  resolves to `Counter`, anchored on the CamelCase boundary (`Discount`, `Account`
  untouched).
- **No unused contract local** in the scaffolded `prePromptModifyContract()`
  (`BPLocalVariableNotUsed` on every captured report).
- **No invented raw-text Labels.** A Label slot takes a label ID or nothing; prose
  titles the RDL only, and the tool says how to set a real label. Shipped metadata
  backs the omission (39/200 sampled TempDB tables, 3/400 `AxMenuItemOutput`).

## Goldens

Five controller goldens are updated in place — the method body has no index
dependency and is pinned by a test. The two report goldens captured from a bare
`fieldsHint` would also change, but those values need the live index and the fixture,
so they stay as captured, are **marked stale in their own READMEs**, and wait for an
`eval-run` re-capture rather than being typed by hand.

`verify-goldens-build.ts --case <id>` is new. **7/7 clean over 38 artifacts**
(xppc 7.0.7996.33, model `fm-mcp`, after a clean `--baseline`) — which also settles
the open question about the two July-frozen goldens: neither goes near the report
scaffold.

## Coverage prose

37 taxonomy notes still claimed "golden pending VM capture" for goldens captured
months ago, one of them published in `COVERAGE.md`. Corrected, and
`tests/eval/coverage.test.ts` now fails when a note and the catalog disagree.

## Five new cases (all `golden_pending`)

`L2-data-types-conversions`, `L2-declarations-scope`, `L2-operators-precedence`,
`L2-statements-switch-loops`, `L2-intrinsic-functions` — the five highest-weight
closure-queue leaves that were taught and validator-backed but proven by nothing.
Each is built around the trap the leaf exists to catch, not around "it compiles".
Coverage deliberately does not move: E comes from a captured golden, never from an
authored spec. Capture is running next; goldens land on this branch.

`docs/XPP_LANGUAGE_COVERAGE_PLAN.md` is removed — the plan is executed and the
durable record lives in `eval/README.md`, the goldens' READMEs and `COVERAGE.md`.

Tests: **5190 passed / 2 skipped, 360 files**.

https://claude.ai/code/session_01DecGW29iVW9UTJbSRmYEb1


dynamics365ninja:
**Update — the five new cases are captured, so this PR now closes the core closure queue.**

All five ran on the VM through the grounded tool path (sandbox `fm-mcp`, prefix `Con`,
xppc 7.0.7996.33): build 0 errors, xppbp 0/0, golden self-match, one corpus record each,
sandbox rolled back. **Core coverage 90.2% → 100% (51/51)**, total 98.9% (88/89) — the one
leaf left in the queue is `attributes` (tier total, weight 2).

Three of the five taught something, and their instructions now carry it:

- **`L2-declarations-scope`** asked for the second loop *after* the first, which proves
  nothing: two sibling `for` loops may legally reuse a counter name. Only a NESTED block
  collides, so the golden nests it. A capture following the original wording would have
  compiled while exercising nothing.
- **`L2-operators-precedence`** built clean but failed xppbp with `BPXmlDocMalformed` — the
  remark described the trap using the operators themselves, and a doc comment is parsed as
  XML, where a bare ampersand is not well-formed; pre-escaping does not survive the write
  path either. Also recorded: the C# bridge refused that `replace-code` as "oldCode not
  found" while the TypeScript writer matched the same text and applied it.
- **`L2-statements-switch-loops`** failed its first build on *"A reference to
  'Dynamics.AX.Retail' is required"* — a `SalesTable` buffer drags Retail's extension of
  that table in. The loop is a container walk now: switch, break scope and flag all survive,
  and the case stays on its own leaf instead of widening the sandbox descriptor.

Every golden README is marked **PENDING HUMAN REVIEW (§6.4)** — a golden becomes a contract
when a person has read it, not when it compiles. That review is the one thing left here.

Tests: 5200 passed / 2 skipped.

dynamics365ninja:
**Update 2 — the closure queue is empty: core 100% (51/51), total 100% (89/89).**

`L2-attribute-authoring-reflection` is authored and captured: a `SysAttribute` subclass, a
class decorated with it (suffix omitted, literal argument) carrying one `SysObsolete` method,
and a reader that walks the attributes back through `DictClass` reflection. Build 0 errors,
xppbp 0/0, golden match, corpus record, sandbox rolled back.

Two findings it now carries:

- xppbp rejected a two-argument `SysObsolete` with `BPCheckSysObsoleteAttributeParametersMismatch`.
  `SysObsoleteAttribute.new` is `(str, boolean, date)` with all three defaulted, and BP still
  wants all three at the usage site — and attribute arguments are positional, so the date
  cannot be skipped.
- `DictClass` and `Array` are kernel types with no AOT metadata row: `get_object_info` says
  "not found" while the code using them compiles. `SysDictClass` does have one, and its
  shipped `getAllAttributes()` source is where the `lastIndex()`/`value()`/`as` walk came
  from rather than a guess.

**Two defects fixed along the way**, both found by using the tools rather than by reading them:

- `eval:score --actual-dir` matched only `*.metadata.xml`, but a live AOT folder holds plain
  `.xml` — so pointing it at `<Model>/<Model>/AxClass`, the obvious thing during a capture,
  paired nothing and scored every artifact `missing`. A silent `golden_match: 0`, not an
  error. `resolveActualFile` accepts a bare `.xml` now (a `.metadata.xml` neighbour still
  wins) and `buildActualArtifactsMap` keys it in the golden filename shape — without that
  second half the resolved pair *still* diffed as missing + extra.
- `npm run lint` was failing on the ASCII assertion added with the report tests: biome's
  `noControlCharactersInRegex` rejects a `\x00-\x7F` range. It is a codePoint filter now,
  which also names the offending characters when it fails.

Tests: 5211 passed / 2 skipped. Lint clean.

PR #1004, opened 2026-09-02
feat(knowledge): H4 catalogs — six gaps, each written from a census instead of an inventory

dynamics365ninja:
H4 of the X++ coverage plan: **G-10, G-12, G-14, G-15, G-17, G-18**. Two new
knowledge entries, four extended ones, one new oracle, one probe batch.

## First: a defect in the censuses H2 and H3 already shipped

They enumerated `<root>/<package>/<package>/<AxType>`. That looks right and
**silently skips the 12 model folders not named after their package** —
`ApplicationSuite/Foundation` among them, one of the largest in the install. The
repo's own `aotSource.ts` walks models correctly and was there the whole time.

`scripts/oracles/usageCensus.ts` now wraps it, `atlNodes.ts` was rewritten onto
it, and **every published figure was re-measured**. The conclusions held; the
numbers moved slightly (the SysTest population is 884 classes, not 488) and are
corrected in the entries, the op-spec, the tests and the taxonomy.

## G-14 — `form-runtime-api`

Two oracles, because the API is split in two and the plan said so. `FormRun` is
an ordinary AOT class (209 methods, member oracle); `xFormRun`, `FormDataSource`,
`FormDataObject` and every `Form*Control` are **kernel with no AOT XML** and only
the compiler can confirm them (probe `coverage-v4h`: 9 of 10 compile, negative
control fails).

The census over all **9,442 shipped forms** is what makes it a catalogue worth
having:

* of FormRun's 209 methods, only **49** are ever called through `element.`, and
  **167 are never called at all**;
* `args()` alone is **21,009 of the 22,913** platform calls — 92%;
* and it **removed** something. `updateDesign` ranks *first* in the raw counts
  (1,703 uses, 724 forms) and is not form-runtime API at all — it is the
  inventory-dimension convention. `UpdateDesignMode` appears in **zero of 76,196**
  shipped files. Breadth of use is not evidence of being platform.

## G-15 — `data-entity-methods`

Ranked over all **5,805 shipped entities**, every signature read from shipped
source. Two things the usual presentation gets wrong:

* `mapEntityToDataSource` and `mapDataSourceToEntity` are **not** a symmetric
  pair — 1,116 vs 99, an 11-to-1 split toward the write path;
* `findEntityDataSource` (318) and `getDefaultingDependencies` (233) are absent
  from the plan's list and beat half of what is on it.

`PrimaryCompanyContext` is an enum with five shipped values, not a boolean.

## G-10 — CoC target kinds

The plan names eight. **`queryStr` has zero shipped uses** — not one of the 4,015
classes carrying `[ExtensionOf]` — while `mapStr` and `viewStr`, which the list
omits, both work and both ship. What `this` *is* differs per kind (buffer, entity,
FormRun, FormDataSource, control), which is the whole reason they are not
interchangeable.

## G-12, G-17, G-18

* `IncludedColumns`: **zero** occurrences in 18,377 shipped tables.
* Relation types censused (Association 15,332 · Composition 1,111 — a
  delete-cascade statement, not a naming preference); the
  `AllowEdit` / `AllowEditOnCreate` split, the two most-set field properties.
* The two SysOperation interfaces declare exactly one method each; nine
  `NumberSeqScopeFactory` creators read off the class.
* All eleven workflow handler signatures — and **`WorkflowQueueCreatedEventHandler`
  is a class, not an interface**, so `implements` against it does not compile.
* The XDS "API" is **two static methods** on `SysSecXDSServices`, confirming H0's
  finding that `XDSServiceBase` does not exist.

## Verification

* full suite **6,127 passed**, 2 skipped, 410 files
* knowledge audit **459 references, 0 defects** — it caught an invented form name
  and a dangling `related:` link before either shipped
* coverage core **72/72**, total **118/118**, closure queue empty
* probe batch valid: the negative control failed, so the "compiles" results mean
  something

https://claude.ai/code/session_01A5usqz9R4hmvptrj4hNw5z

### DIFF
--- a/docs/XPP_LANGUAGE_COVERAGE_PLAN.md
+++ b/docs/XPP_LANGUAGE_COVERAGE_PLAN.md
@@ -26,7 +26,46 @@ Do not keep an executed plan.
 | H3 — G-19 report runtime API, G-23 RDL census | **shipped** | `5618c62` |
 | §6.7 cases — parameters, TDD report-DP, print format, logo+barcode | **shipped** | `c8fe75a`, `27da803`, this branch |
 | **H2 test data & attributes** — G-26, G-27, `atlNodes.generated.ts`, both cases | **shipped** | this branch |
-| H4-H6 | not started | — |
+| **H4 catalogs** — G-10, G-12, G-14, G-15, G-17, G-18 | **shipped** | this branch |
+| H5-H6 | not started | — |
+
+**H4 found a defect in the censuses H2 and H3 had already shipped.** Every one of
+them enumerated `<root>/<package>/<package>/<AxType>`, which looks right and
+silently skips the **12 model folders not named after their package** —
+`ApplicationSuite/Foundation` among them, one of the largest. The repo's own
+`aotSource.ts` walks models correctly and was there all along; hand-rolling the
+walk was the mistake. `scripts/oracles/usageCensus.ts` now wraps it, the ATL
+oracle was rewritten onto it, and every published figure was re-measured. The
+conclusions held; the numbers moved slightly (the SysTest population is 884
+classes, not 488, and `SysTestCheckInTest` is 1,622 uses, not 1,621) and are
+corrected in the entries, the op-spec, the tests and the taxonomy.
+
+**What H4 corrected in its own gap definitions:**
+
+* **G-14.** Of `FormRun`'s 209 methods, only **49** are ever called through
+  `element.` in the 9,442 shipped forms, and `args()` alone is 21,009 of the
+  22,913 platform calls — 92 percent. Publishing 209 would have been a catalogue
+  of things nobody writes. The census also ranked `updateDesign` FIRST (1,703
+  uses, 724 forms) and it is **not form-runtime API at all**: it is the
+  inventory-dimension convention, and `UpdateDesignMode` appears in **zero** of
+  76,196 shipped files. The compiler settled it. Breadth of use is not evidence
+  of being platform — `numberSeqFormHandler`, `enableFields` and `enableButtons`
+  are the same trap.
+* **G-15.** `mapEntityToDataSource` and `mapDataSourceToEntity` are presented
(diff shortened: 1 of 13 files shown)


dynamics365ninja:
## CodeQL `js/regex-injection` — dismissed as won't fix (alert #31)

The 280-character dismissal comment could not hold the reasoning, so it is here.

**Why the finding does not apply.** `scripts/oracles/usageCensus.ts` is a developer-only CLI oracle, run by hand on the build VM. It is not imported by `src/`, not emitted into `dist/`, and not reachable from the MCP server. A regular expression *is* the input this tool takes — it exists so a developer can ask the shipped AOT corpus an arbitrary question — so there is no privilege boundary for a pattern to cross, and escaping the pattern would make the tool do nothing at all.

**Option 3 was considered and rejected.** The only way to break the taint CodeQL tracks is to stop taking the pattern from `argv`, which means cutting the oracle down to a fixed set of queries. Its whole value is that it can be asked anything: H4 exists *because* it could be pointed at `element\.(\w+)\s*\(`, `ExtensionOf\s*\(\s*(\w+)`, `i:type="(AxForm\w*Control)"` and a dozen more on the spot. That is not worth trading for a green tick.

**The alert was not dismissed on sight.** Reading the flagged line properly found three real defects, all fixed in 9cbfcad:

- the regex was constructed **inside** the collect loop, so a sweep recompiled it per file and, under `--declaration`, per method;
- an invalid pattern threw after minutes of scanning, with a stack trace instead of a sentence;
- a nested quantifier backtracks exponentially, and this walks 2.5 GB of XML — `(\w+)+` is easy to type and hangs the run with no output.

The pattern is now compiled once, before the walk, behind a length cap, a nested-quantifier check and a `try/catch` that reports the problem in one line. Both guards were exercised, and the SysTest census returns byte-identical counts (884 files walked, 339 with a hit, `SysTestCheckInTest` 1,622/244).

Dismissed with the repository owner's explicit approval.

https://claude.ai/code/session_01A5usqz9R4hmvptrj4hNw5z

Issue #979, opened 2026-09-01
get_object_info(objectType="form") hides the children of a <DataGroup>-bound group — reports 14 controls where the XML has 16


dynamics365ninja:
Recorded first in `eval/corpus/runs/2026-08-31T22__L3-enum-field-form-downgrade-guard__278eee3.json` (finding 3) and confirmed unfixed by the 2026-09-01 re-run of the same case (`eval/corpus/runs/2026-09-01T06__L3-enum-field-form-downgrade-guard__0fba518.json`).

## What happens

`get_object_info(objectType="form", name=…)` does not list the child controls of a group whose binding is `<DataGroup>Overview</DataGroup>`. In both runs it reported the group `Overview` with **no children** and a total of `Controls: 14`, while the form XML on disk held **16**.

## Why it matters

The reader is the tool an agent uses to find out what a form currently contains, and this is the one group where the answer decides the next write. The eval case's own instruction routes the agent through exactly this call to read the group back — so the agent is told the group is empty, and the natural next step is to add a control that is already there. It also hides #977 (the scaffold writing explicit children under such a group) from anyone inspecting the form through the server rather than by opening the file.

A count that disagrees with the document is worse than a missing feature: it is a number a caller can act on.

## Repro

1. Scaffold or open a form with a group bound via `<DataGroup>` (e.g. `SimpleListDetails` over a table with an `Overview` field group).
2. `get_object_info(objectType="form", name=…)` → the group lists no children, and the total is short by exactly the number it omitted.
3. Count `<AxFormControl>` in the AxForm XML.

## Suggested area

The form reader (`src/tools/read/getObjectInfo.ts` form path and whatever the bridge returns for a `DataGroup`-bound group). Either list the children the document holds, or state explicitly that the group's members are compiler-generated — but the total must match the document either way.

## Related

- #977 — the scaffold writes those children in the first place, in a shape the `add-control` op-spec refuses.

dynamics365ninja:
Fixed in #984 — but the diagnosis in this issue is inverted, and the correction matters more than the fix.

**`get_object_info` was telling the truth. The scaffold was writing controls the platform cannot see.**

The SimpleListDetails template emitted, on its `DetailsHeader > Overview` group:

```xml
<Name>Overview</Name>
<Type>Group</Type>
<DataGroup>Overview</DataGroup>       <!-- both of these belong AFTER <Controls> -->
<DataSource>MyTable</DataSource>
[...]

PR #1008, opened 2026-09-03
fix(report-design, validator): the four findings of the 1.16.2→HEAD release audit

dynamics365ninja:
A release audit of `v1.16.2..main` (9 PRs, 39 commits, 20 runtime files) sent the `src/` diff through an independent review. CI was green, the 6,169-test suite was green, the package was sound — and the headline feature was silently wrong, two new validator rules fired on legal code, and seven of the nine PRs had no CHANGELOG entry. This PR fixes those four findings; the release PR for 1.17.0 stacks on it.

## 1. `add-parameter` wrote `DataType` where the deserializer drops it

The writer emitted `Nullable, PromptString, DataType`. A census of **all 13,911 parameters in the 1,063 shipped reports** puts `DataType` between `AllowBlank` and `Nullable`:

| pair | shipped order | contradicting instances |
|---|---|---:|
| DataType vs PromptString | DataType first | **0** of 3,104 |
| DataType vs Nullable | DataType first | **0** of 2,139 |

The repo's own established finding (#979/#989, `verifyFormElementOrder`) is that an out-of-sequence element is dropped without a word. So the build stayed green and a `System.DateTime` parameter reached the dialog as a `String` — **in the committed golden of `L4-ssrs-report-parameters`, too.** The golden is corrected by hand and its README says why the reviewed file and the capture date disagree; the writer's tests pin the order against what the scaffold itself emits (the scaffold had it right all along). The `axreport-anatomy` topic carries the census, and no longer claims the operation does not exist.

## 2. `add-parameter` wrote half a parameter and called it success

The declaration was written **before** the dataset was resolved. With several datasets and no `datasetName`, a wrong name, or a self-closing `<Parameters />`, the reply was ✅ *"bound it to dataset '(none — no dataset resolved)'"* — and the retry with the right name hit the "already declared" guard, so the parameter could never be bound. The exact half-write the operation's docblock says it exists to prevent.

Now: both insertion sites are resolved before the first byte is written; an unresolved dataset is an error that writes nothing (asserted byte-for-byte); a parameter the old version left declared-but-unbound is **bound on the next call** rather than refused. Also in the same file: `promptString` is XML-escaped, `parameterDataType` must look like a .NET type name, inserted elements take the closing tag's line instead of doubling its indentation, and `refresh-dataset` drops the memoised table listing before the lookup, so a temp table created a moment ago is no longer "not found on disk".

## 3. ATTR003 fired on a legal class-level stack

`maskStringsAndComments` keeps a comment's opener and blanks the rest, so a `///` line comes back as `//` + spaces. The `startsWith('///')` skip therefore never matched, `//` and `/* */` lines were not skipped at all, and the comment line became the "declaration" — which is not `class`, so a legal stack was reported as a parse error. The 105,686-file sweep was clean only because shipped code puts its doc comment *above* the attributes.

TST003 had the same skip and silently did **not** check a test method behind a comment (a false negative, the mirror image). One masked-line helper serves both now; the ATTR003 tests cover `//`, `///`, single- and multi-line block comments, and assert the skip did not become an exemption.

## 4. TST002 could not read a wrapped class header

`([^\n{]*)` stopped at the newline, so `class X` ⏎ `extends Y` was "extends nothing". The header now runs to the opening brace.

## CHANGELOG

PRs #1001–#1007 had no entries under `[Unreleased]`. They do now, each with its measurement, and the dependency bump is under `Changed`.

## Verification

| check | result |
|---|---|
| typecheck, lint | clean |
| full suite | 413 files, **6,190 tests** (21 new), green |
| `eval:knowledge-audit --capture` on the VM | 484 refs, 0 defects, snapshot re-written |
| full-install validator sweep | running, result posted as a comment |

https://claude.ai/code/session_0199Cjao19eyyigVAfeSpfPF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,92 @@ those are called out explicitly below.
 
 
 ### Added
+- **`report-design` — the first write path an AxReport has ever had.**
+  `d365fo_file(action="modify", objectType="report", operation="report-design")`
+  with `reportAction="refresh-dataset"` (copy the temp table's fields onto the
+  report's dataset — the information is on disk, nothing about it is a design
+  decision) or `reportAction="add-parameter"` (declare a parameter AND bind it
+  to the dataset in one write, because the two elements must agree). New
+  parameters: `reportAction`, `tableName`, `datasetName`, `parameterName`,
+  `parameterDataType`, `parameterHidden`, `promptString`. Layout stays with the
+  Report Designer: there is deliberately no `add-column`, because an RDL error
+  surfaces only in the SSRS renderer where no build and no test can see it. The
+  operation is metadata-only and additive-only, and that guarantee is a property
+  of the operation, not of the file's provenance — a fingerprint that refused
+  "foreign" designs was rejected for having the wrong shape.
+
+  It paid for its schema bytes with a measured trim: tool annotations that only
+  repeat the MCP spec default (`readOnlyHint: false` on 6 tools,
+  `idempotentHint: false` on 4) are no longer sent. `destructiveHint: true` is
+  also a default and is kept on purpose: a client that reads absence as
+  "unknown" becomes MORE cautious about a missing `readOnlyHint` and LESS
+  cautious about a missing `destructiveHint`, and only the first direction is
+  safe. ListTools headroom 49 → 251 chars. (PR #1001)
+
+- **`add-field` can write a container field.** `fieldBaseType="Container"` was
+  refused as a parameter that changes nothing and `fieldType="Container"` as a
+  missing EDT; the C# bridge had handled it since it was written and only the
+  TypeScript side never routed to it. A census settled the type question on the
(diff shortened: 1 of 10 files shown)


dynamics365ninja:
## Full-install validator sweep, after the rule changes

`npm run oracle:sweep` over the whole install, detached on the VM (2026-09-03 09:16 UTC):

| | |
|---|---:|
| files swept | 105,686 (615.3 MB of X++) |
| wall time | 3,372 s |
| **error-severity findings** | **0** |
| TST003 (warning) | 10 — unchanged from the #1005 baseline |
| ATTR003 / TST001 / TST002 | 0 |

The bar is unchanged (zero error-severity), and TST003's count is identical to the one #1005 shipped with, so the comment-skip fix did not start flagging shipped tests it previously missed. `lintXppSelect` still names the same 2 advisory files (`BatchRun`, `LedgerJournalEngine_RCash`).

PR #1019, opened 2026-09-09
Fix EDT where-used and CoC extension detection in the xref bridge

kklyandev:
## Issue

`find_references` returned nothing for EDTs. Two independent causes: the
bridge's bare-name path expansion had no `/Edts/` container (nor `/Maps/`,
`/Reports/`, `/MenuItem*/`), and the fallback that then ran had no branch for
`targetType: "edt"` — so it reported `0` without running a query, while
labelling the bridge "unavailable" when it had in fact answered.

That surfaced a larger defect. `CrossReferenceService` documented xref `Kind 2`
as `DerivedFrom/Extends`. It is the generic type reference — 18.3M of 28M rows.
`FindExtensionClasses` was built on that reading and reported every class that
merely mentioned the base as an extension of it: 259 for `SalesFormLetter`
against 9 real, each with fabricated wrapped methods.

## Fix

- Add the missing target containers. `EcoResDescription`: 0 → 261.
- Correct the `Kind` mapping (2 type reference, 3 implements, 4 extends,
  10 override), constant across three xref DBs and two platform versions.
- Detect CoC extensions from the `[ExtensionOf]` declaration: the attribute
  takes one string, the intrinsic emits one reference per metadata level it
  names, and the extended element is the terminal of that prefix chain.
- Group by that element. `SalesTable` is a table (14), a form (17) and 12
  nested elements (17) — previously pooled into a single count.
- Scope wrapped methods to the element; the `SalesTable` form has nine data
  sources carrying an `active` method.
- Drop the hardcoded `Uses 'next' keyword` on the bridge path. The index path
  keeps it — that one actually reads the method body.

### DIFF
--- a/bridge/D365MetadataBridge/Models/Models.cs
+++ b/bridge/D365MetadataBridge/Models/Models.cs
@@ -696,6 +696,19 @@ namespace D365MetadataBridge.Models
         [JsonPropertyName("module")]
         public string? Module { get; set; }
 
+        /// <summary>
+        /// The AOT path this class actually extends, read from its [ExtensionOf] declaration —
+        /// e.g. "/Tables/SalesTable", "/Forms/SalesTable/DataSources/SalesLine",
+        /// "/Forms/VendOpenTrans/DataSources/TaxWithholdTrans/DataFields/TaxReimbursement_IT".
+        ///
+        /// Without this the caller cannot tell a table CoC from a form CoC when both objects
+        /// share a name: asking about "SalesTable" legitimately matches 14 extensions of the
+        /// TABLE and 34 of the FORM and the elements below it, which must never be pooled into
+        /// one count.
+        /// </summary>
+        [JsonPropertyName("extendedElement")]
+        public string ExtendedElement { get; set; } = "";
+
         /// <summary>Methods that the extension class wraps via CoC (next calls)</summary>
         [JsonPropertyName("wrappedMethods")]
         public List<string> WrappedMethods { get; set; } = new List<string>();
--- a/bridge/D365MetadataBridge/Services/CrossReferenceService.cs
+++ b/bridge/D365MetadataBridge/Services/CrossReferenceService.cs
@@ -94,6 +94,24 @@ namespace D365MetadataBridge.Services
         private const byte KindOverride = 10;   // super():            Foo/Methods/pack -> Base/Methods/pack
 
         /// <summary>
+        /// Escape the LIKE metacharacters in a literal that is being CONCATENATED into a LIKE
+        /// pattern. Parameterising the pattern stops SQL injection but not this: inside a LIKE,
+        /// "_" still means "any single character".
+        ///
+        /// That matters here because the D365 CoC naming convention is "&lt;Base&gt;_Extension",
+        /// so essentially every class name this service handles contains an underscore, and an
+        /// unescaped "/Classes/Foo_Extension/Methods/%" also matches "/Classes/FooXExtension/…".
+        /// A census of this xref DB found no name pair that actually collides that way, so this
+        /// is closing a latent hole rather than a live miscount — but the patterns are built from
+        /// class names and element paths, and nothing stops the next model from shipping the pair.
+        ///
+        /// "[" must be replaced FIRST, or it would go on to escape the brackets introduced by the
+        /// other two replacements.
+        /// </summary>
+        private static string EscapeLike(string literal) =>
+            literal.Replace("[", "[[]").Replace("%", "[%]").Replace("_", "[_]");
+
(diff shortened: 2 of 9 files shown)


dynamics365ninja:
Thanks for this — I audited it before merging, and rather than take the empirical claims from the description I went at them against the live `DYNAMICSXREFDB` on my dev box. Different database, different platform build from the three you sampled. Almost everything reproduced.

Two follow-up commits are pushed straight onto your branch (`maintainerCanModify` was on and it seemed friendlier than a stacked PR). Say the word and I'll pull them back out into their own PR.

## What I re-verified independently

| Claim | Result on my xref DB |
|---|---|
| `Kind` 3 = implements, 4 = extends, 10 = override | ✅ Sampling agrees — `*Test → SysTestCase` is 4, interfaces are 3, same-named base method is 10 |
| `Column` is `smallint`, `Kind` is `tinyint` | ✅ `GetInt16`/`GetByte` match the schema exactly |
| `[ExtensionOf]` targets form a prefix chain, zero ambiguous | ✅ 3,403 classes with targets, **0** with more than one terminal |
| The co-attribute column bound is insurance, not a live fix | ✅ Confirmed — removing it changes no answer here either |
| The old query massively over-reports | ✅ `SalesFormLetter`: **304 → 7** (you measured 259 → 9) |
| Dropping `src.Path LIKE '/Classes/%'` is safe | ✅ All 3,031 `[ExtensionOf]` sources are `/Classes/` |

Query time is 2.8 s against the old 4.4 s, comfortably inside the new 60 s timeout.

Worth admitting: my first pass at the chain invariant reported 22 ambiguous classes, which would have sunk the whole approach. That was a missing `DISTINCT` in *my* query, not a hole in yours. Your invariant holds.

## Follow-up 1 — 795e568

`describeUnsearchableType()` treated all three bridge outcomes alike, so an EDT/form/query/view/report where-used that the bridge had *answered* with no rows came back `inconclusive — this is NOT a count of zero`, and advised the reader to "re-run once the xref bridge is available" at the moment the bridge had finished answering them. That is the same mislabel this PR exists to remove, one layer down.

`'empty'` is authoritative by contract — `tryBridgeReferences` returns it only when every query ran and none errored — and it is authoritative *for these five types specifically because of your container-list fix*: the bare name now expands to `/Edts/`, `/Forms/`, `/Queries/`, `/Views/`, `/Reports/`, so the query reaches the rows that exist. So it now reports a real zero, the way the label path in the same file already reports its own, and keeps the inconclusive wording for `error`/`unavailable`.

Every existing "inconclusive" test used an unavailable bridge, so the unsearchable-type + answered-empty combination had no cover. 6 of the 7 added tests fail without the change; the 7th is the guard that `error`/`unavailable` still refuse to give a number.

## Follow-up 2 — 3175773

**LIKE escaping.** Parameterising a pattern stops injection but not wildcards — `_` inside a LIKE still means "any single character", and `<Base>_Extension` is the CoC naming convention, so nearly every class name here carries one. You already escaped `baseClassName` in one spot; that expression is now a named helper, applied to every pattern built from a name or path. The same hole was open in three more places, including the child-path expansion in `FindReferences` — the EDT where-used path this PR adds — plus `FindEventHandlers` and `FindApiUsages`. Values bound to `=` or `IN` are deliberately left raw.

Measured, not assumed: `[_]` matches only a real underscore while bare `_` does match an arbitrary character, and across all 3,031 `[ExtensionOf]` class names the escaped and unescaped patterns select **identical rows — 0 classes affected, 0 extra rows**. So it closes a hole rather than fixing a live miscount, and no current answer moves.

**Counting.** `formatCocExtensions` skipped duplicate class names *inside* the render loop, after `Found N` and the group heading had already counted them. The bridge keys results by class name so none can arrive today, which left the skip either dead code or a guarantee the numbers lie. Dedupe now runs once up front and both counts derive from the list that actually gets printed.

Also re-indented the extension loop body, which stayed at the outer level when the grouping loop was wrapped around it.

## Two notes

- `bridge/build-attestation.json` is regenerated from a real compile (0 warnings). It now records metamodel **7.0.7996.33**, the version on this machine, rather than the 7.0.7996.102 of your build host. Informational field; the gate only checks `sourceHash`.
- Unrelated to this PR, but it caps what the feature can ever return: **423 of the 3,454** `[ExtensionOf]` declarations on my DB have a `SourceId` with no row in `Names`, so any query that inner-joins the source cannot see them. Dangling source ids are widespread in that database generally (200k+), so I read it as the state of my xref DB rather than anything in the code. Flagging it in case you see the same.

Nice piece of work — the `Kind 2` misreading was doing real damage and the `[ExtensionOf]` reconstruction is the right way out of it.

PR #1026, opened 2026-09-17
Fix/stale runtime metadata

laeliand:
The mcp sourced build did not correctly track and contain all changes in the binary files even after a succesful build. Workaround is to clean the xppmeatadata folder and force a full build of objects.

removeModelCompilerMetadata deletes <compilerMetadataPath>\<Model>\XppMetadata before xppc runs, so the write-back phase has nothing to consider current and regenerates it. Since RuntimeMetadataWriter.WriteAll serializes compiler metadata into the binary .md manifests, a stale tree was producing runtime metadata that contradicted the compiled IL — with a green build at every step.

Gated on useFullBuild, which is already fullBuild && modelName === targetModel — so dependencies and incremental builds keep their baseline, which is what -incremental needs.

The call sits after all assertSafePath guards, so the path it deletes is validated.
Missing directory returns early; a failed delete warns instead of failing, and the warning says the .md may not match the code. Both right — the failure mode is the pre-existing behaviour, not a worse one.

rm/access are already imported. tsc --noEmit is clean.


dynamics365ninja (commented):
## Audit

**Verdict: the diagnosis is correct and the fix works, but it covers only half of the problem.** CI is green, and `tests/tools/buildProject.test.ts` and `runBpCheck.test.ts` pass on this branch (107/107).

### Reproduced on a UDE VM (xppc 7.0.7996.33, a small sandbox model, same flags this tool passes)

| Scenario | Result |
|---|---|
| Full build, no clear, sandbox as found | All 24 XppMetadata XMLs left untouched. The source has 1 table, so 23 are phantoms of deleted classes and tables. Even the one real element was not rewritten. |
| Field removed from a class, full build, no clear | Field still declared in XppMetadata ❌ |
| Same, `-incremental` | Field still declared ❌ (**not covered by this PR**) |
| Full build with `-fromVS` | Still stale, so no switch explains it |
| `RuntimeMetadataWriter.WriteAll` over the stale tree | `<Model>_AxClass.md` declares the removed fields, as the PR says |
| **Clear + full build (this PR)** | XppMetadata and `.md` both clean (`.md` 132 → 85 bytes) ✅ |
| Clear + full build **with a compile error** | Metadata is still written back for every element ✅, so a failed full build does not leave the model without metadata |
| Clear + **incremental** build | Only the changed element is written back, and unchanged elements disappear. Gating on `useFullBuild` is therefore correct and necessary. |

Build time was the same with and without the clear (~12–13 s warm), so the "~6s more" in the doc comment is conservative.

From these runs, the write-back appears to write only files that are *missing*, in both modes.

### Findings

1. **Medium: incremental builds (the default) still produce a stale `.md`.** Suggest a follow-up that deletes per-element XppMetadata files before an incremental build, but only these:
   - elements whose source no longer exists;
   - elements whose source is newer than their metadata file.

   xppc does write back a missing file for a changed element (verified). A blanket clear is destructive in incremental mode (see the last row above). Until then, the `fullBuild` description or `incrementalScopeCaveat` could say that only a full build guarantees a correct `.md`.
2. **Low: `removeModelCompilerMetadata` has no test.** The existing CHE test ("never touches the framework directory…") runs incremental only. Suggested cases:
   - `fullBuild: true` → `rm` is called with `<modelStore>\<Model>\XppMetadata`;
   - incremental build → `rm` is not called for it;
   - a dependency in the queue → `rm` is not called for it;
   - `rm` rejects → WARN is logged and xppc still spawns.
3. **Low:** consider `maxRetries` on `rm` (Windows EBUSY/EPERM). A partially deleted tree is no worse than today, since missing files get regenerated, so the comment's reasoning holds.
4. **Info:** if the build is killed with `force: true` between the `rm` and the write-back, the model has no compiler metadata until the next build, and dependent models will fail to resolve it. The window is small; acceptable.
5. **Nits:**
   - The title is the branch name; something like `fix(build): clear the model's compiler metadata on a full build` would fit the log.
   - The `.gitignore` change is unrelated and not mentioned in the PR description. It is correct, though: in git mode `dataRoot()` is the repo root (`src/cli/copilotFiles.ts:87`).
   - **Missing `CHANGELOG.md` entry under `[Unreleased]`.** CHANGELOG.md asks for one in the same PR as the change.

PR #1032, opened 2026-09-18
feat(readers): serve method bodies from the symbol index when bridge and files are unreachable

laeliand:
A read-only deployment has neither a C# bridge (.NET Framework) nor a PackagesLocalDirectory, which are the only two sources the body readers consulted. So every request for X++ source there answered "not found" or "signatures only" — for bodies the server had already indexed.

This allows more functionality for azure based MCP to read and traverse actual source code and not only the references.

symbols.source is populated at index time for every method-owning type and, until now, read back by exactly one caller, on the write path (resolveReferences.findMethod). Measured on a production-size index (2.0 GB): 620,851 method rows, 100% with a non-empty source, averaging 763 characters against the 331-character source_snippet preview.

Adds utils/indexedMethodSource.ts and wires it in as a LAST fallback, so the bridge and the on-disk XML — both of which see newer state than a pipeline-built snapshot — still run first:

  • get_object_info options:{method,include:"source"} — index after bridge and
    XML, and in the inherited-ancestor walk too
  • get_object_info compact:false — the 'source-unavailable' class listing now
    renders bodies, truncated to the same ceiling as the on-disk path
  • find_references — extracts context from the whole body rather than the
    ten-line preview, so a matched row whose call sits further down is no
    longer dropped at the rendering step; scanDeclaringTypeSource falls back
    to indexed bodies when it cannot read the owner's file

No schema change, no re-index, no new storage: the column is already there and already populated. The query plans on idx_parent_type_name (SEARCH symbols USING INDEX idx_parent_type_name), measured at 0.19 ms against the 2.0 GB database.

What this does NOT change: symbols_fts indexes source_snippet, not source, so which methods find_references can FIND is unchanged — a caller that never mentions the name in its first ten lines stays invisible. That needs source in the FTS table, which is a schema change and a re-index.

Bodies from the index are labelled as such in the output, since a snapshot can lag the live model.


dynamics365ninja:
## Audit of this branch

Reviewed at `79fa88a`, measured against this VM's production index (`data/xpp-metadata.db`, 2.5 GB) rather than reasoned about. **The premise holds and the feature works** — eight findings, seven of them now fixed on the branch in `9499ba9` and `e4f84f4`.

### The claims in the description reproduce

| Claim | Measured |
|---|---|
| 620,851 method rows, 100% with a body | 639,397 rows (newer snapshot), **100.00%** with a non-empty `source` |
| avg 763 chars vs the 331-char preview | **763.43** vs **331.72** |
| largest body ~175k | **193,632** chars |
| 0.19 ms per lookup | **1 ms** warm; **272 ms** cold on the widest owner (`SalesTable`) — fine, but the cold path is two orders out |

Also checked: `db-maintenance` never strips `source`, so the published database really does carry it; `get_object_info` is in `ALWAYS_TOOLS` and `find_references` is not in `LOCAL_TOOLS`, so both are reachable in `read-only`. End to end with the parser forced to fail and no bridge (an App Service in miniature), against the real index: `CustTable.find` in **6 ms**, `Tax` with `compact:false` in **17 ms**.

### Fixed on the branch

1. **The index could answer a question the live file had already answered, and label it a lie.** `tryXmlMethodSource` returned the same `null` for "could not read the file" and "read it, no such member". After a method is renamed or deleted and before the next index run: bridge → null, XML parses the current file and finds nothing, index serves the old body stamped *"bridge and metadata files unavailable"* — when one had just been parsed. It now reports which of the two happened and the index runs only for the unreachable case, on the declared-member path and in the ancestor walk alike. Deliberately narrow: this is the reader where a reachable file outranks a snapshot. Paths that consult the index first for its speed (`compact:true` listings, `search`) are untouched.

2. **The indexed listing could carry less than the signatures-only listing it replaced.** It printed the body *instead of* the signature, but an AOT body opens with its `/// <summary>`, so the declaration is regularly past the 200-char ceiling — **44%** of `CustTable`'s methods, **55%** of `SalesLine`'s, **75%** of `Tax`'s. On Azure this path is the only one. The signature is now printed above the body, as the on-disk path has always done.

3. **The listing loaded the owner's whole body set to render one page.** `Tax` = 510 methods / **1.39 MB**, `SalesLine` = 808 / 836 KB, against the ~3 KB actually printed, through synchronous `node:sqlite` — event-loop time. `readIndexedMethodSources` now takes the names to fetch. It also makes the footer hint describe the page rendered rather than the class, which was a second (reported) inconsistency.

4. **The 8 MB / non-file guard in `scanDeclaringTypeSource` had become a no-op** — a refused read fell through to the index and spent the same cost against a different source. "Refused" skips the owner now; only "unreachable" uses the index.

5. **Indexed bodies were concatenated before scanning.** The rows come back in no order, so a call on a body's *first* line took its leading context line from the closing brace of an unrelated method — source that exists nowhere. Each body is scanned on its own.

6. **`readIndexedMethodSource`'s comment named the wrong index.** Measured, both queries plan as `SEARCH symbols USING INDEX idx_type_parent (type=? AND parent_name=?)`, not `idx_parent_type_name`: that index's third column is BINARY, so a `COLLATE NOCASE` comparison cannot seek on it. The conclusion the comment drew was right — the NOCASE compare stays bounded by one owner's methods, never all 639k — but for a different reason, and the test that "verified" it only grepped the SQL string.

7. **The `find_references` half shipped untested.** Five `bodyOf` call sites and the new index fallback had no coverage, and the existing stubs carry `source_snippet` only, so `bodyOf` never took its new branch. Three suites added: the query plans (asked of a real schema, not asserted about a string), `bodyOf`, and the body-boundary splice. All of them fail against the unfixed code.

8. **No CHANGELOG entry**, against the file's own "add an entry in the same PR" rule. Added.

### Checked and refuted, so they need not be re-raised

- **`LIMIT 1` with no `ORDER BY` could pick an arbitrary row** — 400 sampled owners, zero duplicate `(parent_name, method name)` pairs.
- **`parent_name` is ambiguous where a form and a table share a name** (3,000+ such names in the AOT) — for all 3,000, the method rows span exactly one file. No owner-type filter is needed.

### One thing left as a judgement call

`bodyOf` widens the text the extractors scan, and on real data that buys less than it might seem: of 150 FTS-matched rows, rendered references went `validateWrite` 130 → 138, `initFromCustTable` 65 → 67, `parmSalesId` 125 → 126, `findOrCreate` 149 → 149 — **+0 to +6%**. The description's own motivating case, `buildAdjustIn`, still returns 0 rows, exactly as the description says it will: discovery is still bound to `source_snippet`. It costs almost nothing and cannot worsen an existing result (the preview is a prefix of the body, and the extractors take the first match), so it is worth keeping — but the real Azure value in that file is the `scanDeclaringTypeSource` fallback, not the widening.

### Verification

`tsc --noEmit` clean · `biome lint` clean · full suite **6,413 passed / 2 skipped** · every new test confirmed to fail against the pre-fix code.

